PR #1, opened 2024-06-06
OpenVR device managment

CascadianVR:
started work on implementing OpenVR for Basis.


dooly123:
got a few bugs, will be merged once checked :) first one!

CascadianVR:
Fixed some bugs
The openvr package should now be added automatically to the package manager via a local .tgz file.
Headset and controllers now track, connect and disconnect properly.

Fixed a big where opening the steamvr menu freezes tracking.
One known bug is sometimes controller roles (left and right) are swapped or the same. Turning them off and on fixes this for now.

I plan on adding input now for movement.

dooly123:
looks good will like everything else needs some TLC but another problem for another day

PR #5, opened 2024-06-08
Input bug, Avatar config saving, avatar material update

CascadianVR:
Waiting a few frames when initializing input device due to weird steamvr input bug not initializing the first frame. Fixed avatar component not saving configuration for eye and mouth positions. Updated avatar eye material


dooly123 (approved):
yep still needs work but lets go for it

PR #23, opened 2024-09-27
fixed the mirror on desktop

dooly123:
time top test


dooly123:
works in desktop just fine, 


switching into vr borks it. (need to handle camera switching from and too vr.

rotation in vr is also wrong.

PR #26, opened 2024-09-30
Execute Non-Destructive Modular Framework during avatar builds if it is installed in the project

hai-vr:
If Non-Destructive Modular Framework is installed in the project, execute it when building an avatar from a Prefab.

Only when hooked, the Prefab is instantiated, the processor runs, and a new Prefab in created in the Assets/ZZZ_Generated/ folder.

If NDMF is not installed, the build process remains unchanged from the original behavior.

### DIFF
--- a/Packages/HVRBasisNDMF/Scripts/Editor.meta
+++ b/Packages/HVRBasisNDMF/Scripts/Editor.meta
@@ -1,0 +1,8 @@
+﻿fileFormatVersion: 2
+guid: 510e6e3a9936bba449517be785e685e0
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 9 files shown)


hai-vr:
I was able to confirm that non-destructive avatar components such as Modular Avatar Merge Armature and Prefabulous could run during the avatar prefab build process for Basis.

- NDMF itself appears to work mostly; it does have the same general import issues as other apps, as it requires some DLLs to exist in the project first (HarmonyLib, System Collections Immutable)
- As for Modular Avatar, it can run with a few modifications (https://github.com/bdunderscore/modular-avatar/pull/1232). Many components remain unavailable in Basis, but at least [Merge Armature](https://modular-avatar.nadena.dev/docs/reference/merge-armature) works.
- I don't have a solution at the moment for the requirement that local avatar tests must be Addressable prefabs (instead of scene objects), so to test the avatar in-editor, it needs to be built first, and then the generated prefab is to be made Addressable.

Unity_eWW83LCOfy

PR #27, opened 2024-09-30
Distance-Based UI-Reparenting

RadioArtz:
The BasisUIReparenter reparents the UI object to the LocalPlayer if they move too far away from the UI. This solves the issue where the UI becomes unusable if the player falls infinitely or is involuntarily moving in some other way.

While it might be preferable to make the distance a global setting instead of configuring it per script, I've opted not to implement that for now to avoid cluttering the settings UI.

https://github.com/user-attachments/assets/66b50b85-f374-4cff-96ca-3a66e4c98ad9


dooly123:
thanks for taking the time to fix this part.

dooly123:
I have merged this in however i was able to get the UI to flip 180 when moving away, 

let me know if you can replicate.

RadioArtz:
Have also gotten this report from another source, gonna look into a fix!

PR #28, opened 2024-10-13
Add native rendering plugin

Error-mdl:
Added simple native rendering plugin to patch vulkan sampler states to add linear mip blending (aka trilinear filtering) when using anisotropic sampling to match d3d11 behavior in unity. Potentially, this plugin could be used for other things in the future. Source code for plugin will be pushed to a separate repo.


Error-mdl:
source code is temporarily here, should be moved to a repo owned by dooly: [https://github.com/Error-mdl/tempBasisNativeGfx](https://github.com/Error-mdl/tempBasisNativeGfx)

dooly123:
thanks, error means a lot, man my life is getting busy with all this.

PR #42, opened 2024-10-30
Skip overly defensive controller rejection

yewnyx:


dooly123:
still defensive but excluding left and right controller.

PR #45, opened 2024-11-30
Philosophy of Basis

yewnyx:
A first pass at explaining the principles of Basis, especially to avoid unfortunate misunderstandings.


yewnyx:
(note for anyone looking back at this: I am the original author)

PR #50, opened 2024-12-02
:sparkles: Highlight microphone icon when talking

AnotherFoxGuy:
This PR adds a simple highlight to the microphone icon when it receives audio
This should make it way more obvious when your microphone isn't sending any audio

### DIFF
--- a/Assets/Prefabs/Loadins/Main Camera.prefab
+++ b/Assets/Prefabs/Loadins/Main Camera.prefab
@@ -400,6 +400,7 @@ MonoBehaviour:
   MicrophoneCanvas: {fileID: 3373432966423603715}
   MicrophoneMutedIcon: {fileID: 7951072778522726859}
   MicrophoneUnMutedIcon: {fileID: 1908197658278707850}
+  MicrophoneUnMutedIconTransform: {fileID: 1619504432556682231}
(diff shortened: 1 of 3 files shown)


dooly123:
will test shortly, thanks for taking the time and effort it's appreciated!

dooly123:
ok, it will work, I will update the code to in the update loop and skip setting it if the colour hasn't changed. let me know if you want to do that tho.

dooly123:
would fading in and out look better aswell?

AnotherFoxGuy:
>  skip setting it if the colour hasn't changed

I can add that 

> would fading in and out look better aswell?

I was thinking of maybe scaling it up a bit when active, but I'm not sure if that would become annoying after a while

dooly123:
only way to know is to try!

> > skip setting it if the colour hasn't changed
> 
> I can add that
> 
> > would fading in and out look better aswell?
> 
> I was thinking of maybe scaling it up a bit when active, but I'm not sure if that would become annoying after a while

AnotherFoxGuy:
I've added some scaling to the icon, let me know if it is too much or annoying

dooly123:
You all good love committers merging tomorrow

PR #52, opened 2024-12-09
When using content removal, add option to disallow animator events:

hai-vr:
- Add `disallowAnimatorEvents` content policing.
- When enabled, Animator.fireEvents will be set to false on all Animator components found in the asset.


hai-vr:
Just realized, the content policing file is presented as an allowlist "check to permit", but this option is presented as "check to restrict". It may be relevant to invert the meaning so that everything in the content policing is presented as an allowlist

dooly123:
thanks for all the finds hai hopefully the new code holds up

PR #54, opened 2024-12-10
Add DNS Resolve

MidnightWaffles:
This code will select IPv4 if it can find it in the DNS response.

I strongly urge that we move to IPv6 based on modern day hosting costs. But for now this will select the first IPv4 that the hostname resolves to.


dooly123:
thanks will make sure to tackle this for the new server aswell

PR #55, opened 2024-12-11
Make NDMF create an Addressable prefab when entering Play Mode:

hai-vr:
- If we use NDMF, entering Play Mode is not sufficient to equip the avatar for local testing.
- We need to create an Addressable prefab out of the avatar, but only after the avatar has been serialized by NDMF, otherwise we'll get missing references.
- NDMF serializes the avatar only when the build context "finishes", which is after all NDMF passes are completed.
- To hack around this, create a NDMF plugin that runs at the end of the Optimizing phase, which will delay an invocation later which will turn the processed and serialized avatar into an Addressable prefab.
- Only one avatar per scene is supported.


hai-vr:
Sorry, I meant to open this PR on my repo, not upstream, please ignore for now

PR #57, opened 2024-12-12
Fix website markdown rendering

mriise:
Jekyll _may_ need to have a different markdown renderer defined, or its just bad markdown. haven't tested locally, so who knows yet!

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,8 @@
+theme: jekyll-theme-minimal
+title: Basis
+description: Basis Framework
+lsi: false
+safe: true
+kramdown:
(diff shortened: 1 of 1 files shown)


mriise:
it is, in fact- not fixed :)

should not be lazy and test locally lol.

dooly123:
me and you both, where should i look for this?

dooly123:
i googled what it was doing and I'm like "sounds good"

mriise:
its probably some unsupported feature in the markdown for the README. Ima build jekyll locally and fiddle with it till it looks passable.

PR #58, opened 2024-12-12
changed menu background and hierarky

spot-neko:
made background of menu less transparent and made menu text a chiled of the menu setting for the setting text : 3


dooly123:
closed this, redid the commit and set you as the author.

PR #59, opened 2024-12-13
Work towards adding Toggles, and Face Tracking fixes:

hai-vr:
- Work towards adding Toggles:
  - Add Object State Actuation component.
  - Toggling from OFF to ON sends a packet.
  - Toggling from ON to OFF sends a packet.
  - Late joining / Late loading: When a remote player loads the avatar, it asks the avatar wearer to send the state of the toggle to them.
  - Early loading: If a remote player loads the avatar faster than the wearer, the wearer will send the state of the toggle when the wearer finishes loading.
  - Unlike face tracking, this is not a continuous stream of data, so no packets are sent continuously if the toggle remains untouched.
- Improve late loading and early loading in the HVRAvatarComms protocol:
  - When a non-wearer finishes loading the avatar networking, a Reserved packet of type "Remote Requests Initialization" is sent to the wearer, so that the wearer sends a Negotiation packet to them.
  - When the wearer finishes loading the avatar networking, a Negotiation packet is sent to everyone else.
  - Remove periodic Negotiation packet, which was a hack for which the aforementioned changes replace.
- Null values are removed from user-provided arrays (i.e. in BlendshapeActuation).
- Support resyncing in features:
  - When after a Negotiation packet is sent, the wearer also asks the features to submit packets for resyncing state.
- AcquisitionAssist can now feed data with values 0 and 1, presented as buttons to test toggles.
- Face Tracking fixes:
  - Blendshape Actuation is now less sensitive to execution order of OnAvatarReady and OnAvatarNetworkReady, as the order appears to be currently different if the avatar is local or remote.
  - Fix data Acquisition was incorrectly feeding data for remote objects.
  - Eye Tracking Bone Actuation is now a ICommsNetworkable component, as eye bones are no longer networked in Basis. This requires a reupload of avatars, since this component now needs to be registered with a GUID in Feature Networking.
  - Fix Eye Tracking now fetches the eye bone driver of both the local player and remote players, instead of just the local player.
  - Fix execution order is now specified to run after other bone drivers that was interfering with the bone rotation.
  - Add a hack so that eye lids aren't closed by default. We need to add default values in the blendshape definition files.

### DIFF
--- a/Packages/HVRBasisAvatarComms/Scripts/Acquisition/AcquisitionAssist.cs
+++ b/Packages/HVRBasisAvatarComms/Scripts/Acquisition/AcquisitionAssist.cs
@@ -11,6 +11,8 @@ namespace HVR.Basis.Comms
         [HideInInspector] public AcquisitionService acquisitionService;
         [NonSerialized] public Dictionary<string, float> memory = new Dictionary<string, float>();
 
+        public string[] toggles;
+
         private void Awake()
         {
             if (acquisitionService == null) acquisitionService = AcquisitionService.SceneInstance;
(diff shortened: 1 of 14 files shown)


dooly123:
awesome work hai,  ok to merge?

hai-vr:
It should be OK to merge, when I'll have other improvements that's gonna be with another PR

PR #65, opened 2024-12-17
Make menu panels request visibility of trackers:

hai-vr:
- Fix trackers are no longer visible if the user closes the menu using the X button, or by respawning, or by opening the settings, or by opening the avatar menu.
- Basis UI Menus now must request in their UI Initialize event for the trackers to be visible, and un-request in the UI Destroy event.
- Add BasisTrackerVisibilityHandler MonoBehaviour in the InputActions prefab.
  - This behaviour contains a coroutine that is run upon at the end of frame after requests and un-requests are submitted, so that when a menu closes itself to open a different menu, the tracker visibility is only toggled ON or OFF once for any given frame.

NOTE: This PR conflicts with the Snap Turn PR because they both modify the InputActions prefab.

### DIFF
--- a/Assets/Prefabs/Boot/InputActions.prefab
+++ b/Assets/Prefabs/Boot/InputActions.prefab
@@ -169,6 +169,7 @@ GameObject:
   - component: {fileID: 256212325347645450}
   - component: {fileID: 6482417607131076391}
   - component: {fileID: 6378574261833472064}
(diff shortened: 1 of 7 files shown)


dooly123:
i have merged the other feel free to merge in and go for a merge request

hai-vr:
Ready

PR #70, opened 2024-12-19
Fix avatar should no longer jitter when app runs at low framerate:

hai-vr:
- When the app runs at a low framerate, the avatar bone rotations were affected by jitter, both in local and as visibly seen by other users in the network.
- Fix by reusing the same interpolation factor clamping logic from a different location within BasisBoneControl.
- There still seems to be some rotational jitter from the lower body twisting to face forward; this is probably a separate but similar in the character controller. This commit does not fix this.


dooly123 (approved):
thanks hai

PR #75, opened 2024-12-20
When the local user is speaking, increase face tracking send rate:

hai-vr:
- When the local user is speaking, increase the face tracking packet rate to 33 packets per second instead of 10 packets per second.
- This is experimental to see if better mouth movements could be captured.

### DIFF
--- a/Basis/Packages/Basis Framework/Networking/Transmitters/BasisNetworkTransmitter.cs
+++ b/Basis/Packages/Basis Framework/Networking/Transmitters/BasisNetworkTransmitter.cs
@@ -24,7 +24,7 @@ namespace Basis.Scripts.Networking.Transmitters
     {
         public bool HasEvents = false;
         public float timer = 0f;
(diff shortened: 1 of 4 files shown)


hai-vr:
> you should also look at grabbing the interval rate from the networked player's transmitter,

It's now plugged into the interval (using `(interval / DefaultInterval)` as multiplier)

dooly123:
mint, let me know if you can get this in before this weeks test would help with a lot of people

hai-vr:
> mint, let me know if you can get this in before this weeks test would help with a lot of people

This change will make it heavier, not lighter.
- When users are clustered together, it's practically the same rate as the previous load test which is 10 messages per second.
- With this change, whenever a player is speaking through a avatar that supports face tracking, it would be 30 messages per second.
- We are around 92 bytes of data per message (not including headers), so that's going from 920 bytes to around 2760 bytes per second.

PR #78, opened 2024-12-21
Make SkinnedMeshes of the local player head cast shadows:

hai-vr:
- The SkinnedMeshes of the local player that reference any bone under the Head bone should now cast shadows of the head onto the world and other players.
- SkinnedMeshes placed directly under the Head bone without any bone reference are not supported in this version.
- MeshRenderers placed directly under the Head bone are not supported in this version.
- This implementation:
  - Evaluates at runtime.
  - Creates a copy of all the transforms under the Head hierarchy.
  - Creates copies of SkinnedMeshes that reference any bone under the Head.
    - Bones under Head will reference the aforementionned copy.
    - Bones not under Head will reference a zero-scale bone under the neck.
    - Those SkinnedMeshes are set to Shadow Only.
  - This implementation currently uses "Update When Offscreen" OFF and an off-world "Bounds" value to prevent the shadow from displaying when rendering the local player avatar in third person.
    - This implementation was chosen because it's currently unclear how to manage the camera layers, and should be discussed further.
  - Just before any camera is rendered, once per frame, we copy the transforms, SkinnedMesh component enabled-ness, and blendshapes, so that the shadow reflects the visual state of the original SkinnedMesh.

### DIFF
--- a/Basis/Packages/Basis Framework/Avatar/BasisAvatarFactory.cs
+++ b/Basis/Packages/Basis Framework/Avatar/BasisAvatarFactory.cs
@@ -7,13 +7,14 @@ using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
+using Basis.Scripts.BasisSdk.Helpers;
 using UnityEngine;
 using UnityEngine.AddressableAssets;
 namespace Basis.Scripts.Avatar
 {
     public static class BasisAvatarFactory
     {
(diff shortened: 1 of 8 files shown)


dooly123:
is this ready?

hai-vr:
> is this ready?

Is it functional, but ideally I'd like a second opinion from someone who is familiar with shadow "clones", i.e. Lyuma.

dooly123:
@lyuma

PR #79, opened 2024-12-22
Ignore all cursor management requests while in VR mode:

hai-vr:
- When entering VR mode, unlock cursor and ignore all cursor management requests.
- This new behaviour prevents the cursor from being stolen from other VR desktop overlay applications.


dooly123:
Heyo the bool should be swapped for my is in vr check that way it's one spot not two. Should be a helper function somewhere

hai-vr:
Updated to use `BasisDeviceManagement.IsUserInDesktop()`

PR #82, opened 2024-12-24
fix wrong variable name

TheButlah:
Writer -> Reader


dooly123:
bruh breaking my copy paste fun

PR #88, opened 2024-12-27
Correct File Extension on Generated Password File

Enverex:
SaveFileAsync only needs the extension, the period is automatic. This fixes the double period in filename issue.


dooly123:
another day another bug.

PR #81, opened 2024-12-23
delete bin directory

TheButlah:

PR #89, opened 2024-12-28
Separate gitignores

neon-glowstick:
There are two projects in the repo and I think it makes sense to have a separate gitignore for each of them. I kept the gitignore at the root to exclude the server build output. Files are from https://github.com/github/gitignore/blob/main/Unity.gitignore and https://github.com/github/gitignore/blob/main/community/DotNet/core.gitignore. 

Branched from New-Server-Ready because that is where #81 was merged.

I've dropped the ignored files in a [separate branch](https://github.com/neon-glowstick/Basis/pull/1) and it seems to work fine on a clean clone.


dooly123:
just looking over it now, did not even know that git supported multiple git ignores, things you learn  ;)

dooly123:
Had a chat with one or two other people about this, they both don't want it being 2 separate files. (guess it comes down to preference) will run it by the other contributes and see what they think

neon-glowstick:
The argument for multiple files is that you have two separate codebases the same git repo. If you decide to move a project its easier if the gitignore follows along with it. And the projects may have conflicting requirements, such as Basis (the unity project) which should not include .csproj files, but Basis Server (a dotnet project) must include them.

yewnyx:
Strong agree that separate gitignores are preferable for multi project repositories

dooly123:
sounds like we will go ahead and do it then :)

dooly123:
tested all looks good, thanks!

PR #90, opened 2024-12-30
feat(auth): implement did:key resolution

TheButlah:
~WIP, do not merge.~

I implemented [did:key resolution](https://w3c-ccg.github.io/did-method-key). This implements the necessary logic to resolve a did:key to a did document, and extract the public key from that document.

In a subsequent PR, I will actually use this decoded public key to do authentication.


dooly123:
will merge when your at the deepest part of your sleep

TheButlah:
I think I implemented decoding correctly, but I'll add some tests next time I work on this.

TheButlah:
Added a test, it passes. Ready for review.

PR #91, opened 2024-12-30
fix: gitignore common ide files and c# build artifacts

TheButlah:


neon-glowstick:
I branched off this and made #92 to drop all files that should be ignored from the repo

neon-glowstick:
I think its fine to have these generic rules at the root and that it is not a problem that they become duplicated by the more specific rules. The point of having multiple gitignores is to scope rules for the separate codebases. The rules for Basis and Basis Server _should_ be complete on their own so that they continue to work, even if they are moved down the line.

PR #104, opened 2025-01-08
bug/compression-jitter

salmoncheeks:
- Fixed micro jitters of compressed network data for networked muscle data.
- Added OneEuroFilter as Util to be used to smooth out jittery signals.
- Added HumanBoneMotion enum for future use.


dooly123:
thankyou, time to test this at scale :)

PR #114, opened 2025-01-20
Refactor Raycast - Round 2!

mriise:
WIP. do not merge

Need a full review pass to verify behavior is consistent with pre-refactor UI handling.

## Changed
- ignore localplayer in raycast

## TODO
- raycast rotation offset, atm its ignored and we just use game object transform
- make sure that second UI graphic sort is ok to remove


dooly123:
i am going to merge this into doolys branch as its in a better state then whats on there currently. getting close :)

PR #119, opened 2025-01-25
Drop unused meta files

neon-glowstick:
These .meta files are automatically deleted by Unity because the .csproj and .sln files they relate to do not exist in the repo. Delete them to unclutter git on a clean clone.


dooly123:
I won't be merging this, as the files end up inside of unity, regening the meta files every time I do that will lead to issues.

PR #120, opened 2025-01-30
Enhanced Health Check endpoint with Stats when Stats are enabled.

MidnightWaffles:
Now when Server Statistics are enabled you will see more helpful details.


dooly123:
your caching the data even if stats are disabled? think that might need changing also please target doolys branch. nothing gets merged into lts anymore without going from dooly -> main -> lts

PR #140, opened 2025-02-09
build(did): add package to basis along with transitive deps

TheButlah:


TheButlah:
@dooly123 I think this is correct but lmk. Please test before merging if u can :3

PR #147, opened 2025-02-13
Convenience UI for uploading avatars to S3 buckets

neon-glowstick:
## Objective

This PR adds a new Contrib folder for uploads and an implementation for uploading avatars to an S3 bucket.

To use the package add `"org.basisvr.contrib.upload.s3": "file:../../Contrib/Upload/S3"` to manifest.json. 

image

The textfields are **NOT** serialized on the component. They are saved to a config file in [Unitys PersistentDataPath](https://docs.unity3d.com/6000.0/Documentation/ScriptReference/Application-persistentDataPath.html) locally on your machine. Use the Load/SaveDelete buttons to manage the config file. 

I tested it with Cloudflare and a public R2 bucket in dev mode. Upload went fine and the avatar downloaded without issue into Basis.

## Dependencies
I'm using the [AWS S3 .net sdk](https://www.nuget.org/packages/AWSSDK.S3) to do the actual upload. It was downloaded using [NugetForUnity](https://github.com/GlitchEnzo/NuGetForUnity) then copied into the `Contrib/Upload/S3/Editor/libraries` folder. The AWS sdk packages are under the Apache 2.0 license, and the depend on [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) which is under the MIT license.


dooly123:
this is awesome, I think it needs to be on a separate repo however, (just so we don't add addons to the framework)

bring it up with me when you got some time at the meeting or on discord and we can discuss how that would look :)

neon-glowstick:
I've made this into a package here https://github.com/neon-glowstick/BasisVr-S3Hosting
My intent is to add support for uploading scenes and props as well.

PR #157, opened 2025-02-20
Updated SteamVR Input Bindings and Gamepad Bindings

MehStrongBadMeh:
Updated SteamVR input profiles for Index and Meta Touch to fix up the grip function. Created a SteamVR input profile for Vive. Updated Gamepad input profile to fix missing jump and menu commands.

Due to Windows dropping support for WMR, I won't be making WMR or HP Motion Controller profiles.


dooly123:
awesome work :) looking forward to testing it this week

PR #167, opened 2025-02-24
Replace Camera Defaults with more Robust Defaults

Enverex:
Adjust the far-clip to be much further out and adjust the near-clip to compensate, should still be close enough for almost all cases.

Fix the "Quake 3 Arena"-esque default FoV with a more normal looking camera FoV.

Also expose the farClip value so it can be easily adjusted.


dooly123:
The role of BasisSceneFactory is to set the camera settings; as such, I don't see how changing these values is a good idea, especially the near clip, which will cause issues.

I think adding FOV as an option in the settings panel is something that can be done. (edit done so will be on main in the next few days)

PR #169, opened 2025-03-03
feat(Crypto,Did): implement rest of did:key

TheButlah:
Implements everything remaining in did:key, including:
* signing ed25519 cryptography + tests
* Converting from a pubkey to a did:key
* Sample server/client code as a test case in ServerExample.cs
* added the `Generator.Equals` package from nuget and made all records use value equality
* fixed a bug in GetAlgorithm
* Updated unity project to have correct dependencies and settings


TheButlah:
marking as a draft till I can finish my example test case

TheButlah:
ready to merge

PR #203, opened 2025-04-05
fix(msbuild): remove and correct broken project references

TheButlah:
when the `Contrib` directory was moved inside the `Basis Server`, a bunch of old csproj and sln references were broken. This PR fixes it:
* only references to contrib stuff are in toplevel sln
* all other csproj ProjectReferences are using the correct paths now.


TheButlah:
Tried to test but there are a ton of unreferenced assemblies etc. I think `main` is just broken

dooly123 (approved):

PR #204, opened 2025-04-05
feat(handles): dns handles

TheButlah:
**Merge #203 first**

See the README.md files in the files changed for an explanation of how this is structured and what it does.
See Dns.Tests for the example usage.

I tried to make it so that `Handles` (common folder) is a minimal library of just the interfaces and some necessary types, and `Dns` is a package that could live in `Contrib` with a dependency on `Handles`. Once we start thinking about how to integrate this into the demo project, maybe we should put `Handles` into core, but keep dns, steam, etc in Contrib.

But thats a dooly problem 😉


Basis Server/Contrib/Handles/Common/Common.cs
```diff
@@ -0,0 +1,65 @@
...
+		/// Identities can be added and deleted from the set at will.
+		Mutable,
+	}
+
+	/// The different supported DidMethods.
+	public enum HandleKind
+	{
+		Local,
+		Dns,
+		// TODO: HttpWellKnown
```
dooly123, line 62: we should make it so its

Local,
Dns,
Other, 

this way developers can implement there own type without needing to modify this file.

dooly123 (approved):

PR #218, opened 2025-04-23
Pickups Polishing 3

### LINKED ISSUE #136: Pickup local transform offset in vr is incorrect
### Describe the bug?

object will offset in an incorrect direction

### Steps to Reproduce

grab a pickup

### The expected behavior

not offset all weird like

### Screenshots

_No response_

### BasisVR Release Version

doolysbranch

### What VR headsets are being used during this issue?

index

### Log Files

N/A

### Additional Context

This is likely my implementation of parent constraint being incorrect. `BasisParentConstraint`

mriise:
- use bone control for all constraint calculations (fixes #136)
- Add toggle for self-stealing pickups
- document when willInteract is set
- dont set velocity on kinematic objects
- pre-filter interactables by IsInfluencable when getting nearest
- rename confusing "interacting" with "influencing" (either hovering or interacting)
- run format
- cleanup old TODOs


mriise:
There is an issue with hover sorting where it will select the next nearest object when the nearer object just had interact end for a single frame before correcting itself. It needs better documentation and and its own issue when this is merged, i do not plan on fixing it here.

PR #225, opened 2025-04-28
docs: break trademark policy into its own file

TheButlah:
Makes the readme less cluttered while keeping the trademark policy easily findable.

### DIFF
--- a/TRADEMARK.md
+++ b/TRADEMARK.md
@@ -12,6 +12,6 @@ If there is any confusion about these permissions, please get in touch so we can
 
 TL;DR:
 * MIT covers the code; name/logo remain trademarked.
-* Use “Built with Basis” descriptively—no implied endorsement.
+* Use “Built with Basis” descriptively - no implied endorsement.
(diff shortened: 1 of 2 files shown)


TRADEMARK.md
```diff
@@ -0,0 +1,17 @@
...
+
+We explicitly grant permission to use the name and logo for contributions made to the Basis Project, its documentation, or educational materials or tutorials for Basis. Additionally, we authorize use of the name and logo in text or video to truthfully refer to and/or link to the Basis Project and Framework.
+
+If there is any confusion about these permissions, please get in touch so we can help clarify or improve our guidelines.
+
+TL;DR:
+* MIT covers the code; name/logo remain trademarked.
+* Use “Built with Basis” descriptively - no implied endorsement.
+* Don’t name forks using the Basis name i.e. “Basis 2/Pro”, or use the logo without permission.
+* Contact us if unsure.
```
kitlith, line 17: Keeping the TL;DR: in the readme might be useful. Unsure if it should be included in both places or not.

PR #230, opened 2025-05-01
Fixing path in .vrmanifest to point to relative actions.json

Soapwood:
This was pointing to hardcoded local developer path.

@dooly123 Does this make sense?

To be honest this is kind of a quick fix - ideally the `actions.json` should be in the user's appdata when the client is installed through steam, where the .vrmanifest can always reference it. That would be the long term fix imo.

### DIFF
--- a/Basis/unityProject.vrmanifest
+++ b/Basis/unityProject.vrmanifest
@@ -5,7 +5,7 @@
       "app_key": "application.generated.unity.basisunity.exe",
       "launch_type": "url",
       "url": "steam://launch/",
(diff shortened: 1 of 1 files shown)


dooly123:
@MehStrongBadMeh mind taking a quick look ty :)

Soapwood:
> also we normally always make changes to doolys branch first and then it goes
> 
> doolys branch -> main -> lts

Understood, cheers! Will know for next time.

MehStrongBadMeh (approved):
Yeah, that is the correct path

PR #241, opened 2025-05-07
Adds a prebuild bundle event list for locking shadurs

SCRN-VRC:
Waits for all tasks to finish before building


dooly123:
we always merged into doolys branch or main before lts

PR #251, opened 2025-05-18
BasisDistanceJobs: add hysteresis margin to prevent flapping

dakyneko:
Logic used for checks in DistanceBasedReduction may have flapping issues: an avatar could appear and disappear quickly because it's on the edge of the "Avatar range" setting. Towneh in discord reported the issue and I found interesting to solve the issue as a starting point to contribute. This only applies to the demo I think.

I've added a margin around the threshold distance depending on which state we're in: inside or outside it.
Towneh described this as a "grace distance gap between loading and unloading the avatar at the user setting configured distance".
This is a proposal as I haven't compiled the project myself yet (untested code). Feel free to give your opinion on this.


dooly123:
tested this code all looks good, merge approved :)

PR #257, opened 2025-06-06
Remove embedded audiolink

techanon:
Remove audiolink from the framework as that should be added independently by the framework consumer.


dooly123:
im actually hoping to keep it around at least until we have a proper way to remove all the #if other code

techanon:
@dooly123 what other #if code?

dooly123:
im going to close this and remove it from the framework, I'm planning on spinning up a second repo (demo repo this week)

PR #259, opened 2025-06-10
Pickup polishing And Fly Camera

mriise:
**DO NOT MERGE** until #258 lands or a different approach is reached.

im pretty sure @dooly123 had some pickup changes that might be good to wait for to avoid conflicts.

Sorry this PR isnt split better but it just so happened the things i was working on all needed changes at the same time.

## Changes

Pickups:
- better (mostly untested) handling of network pickups and stealing
- max desktop zoop distance
- proper velocity on drop
- use direct transforms setting when Rigidbody reference is null
- dont hover/interact if UI raycast had a UI target
- when non-kinematic pickup disable gravity while held
- only trigger OnInfluenceEnable/Disable if the value changed state

InputModule:
- movement, crouch, jump pause requests (temporary implementation, see #258)

Camera:
- fly movement
- inherit from pickup to avoid repetitive code
- camera pinning (handheld, playspace, world)
- force camera preview to stay in front of face on desktop
- lots of camera settings for cinematic style movement

BasisParentConstraint:
- blend quaternions better (little testing of this change done)
- index gaurds

## TODO
test in vr, havent tried it at all yet.
test net logic


Basis/Packages/com.basis.framework/Networking/BasisNetworkPlayer.cs
mriise, line 1: @dooly123 these ok to merge in? or reset back

Basis/Packages/com.basis.framework/Players/BasisPlayer.cs
mriise, line 1: this too...

PR #302, opened 2025-07-13
Fixed show folder feature for Linux

naelstrof:
`open` isn't a command on Linux, most desktop environments support the XDG standard and use `xdg-open`. However this isn't guaranteed and I think should not cause a build failure if it doesn't exist.


dooly123:
meta not testing on linux, who would guess

PR #305, opened 2025-07-16
CI: fix failing server builds and also build the client for Linux and Android

kitlith:
There was a lot of history rewriting using `jj` involved to get a clean history, but this should be reviewable commit-by-commit. I am marking it as a draft until I confirm that everything is working correctly/as envisioned.

You will need to add the android keystore secrets to your CI setup before you merge this.
Namely, the first 4 secrets mentioned at <https://game.ci/docs/github/deployment/android#7-add-secrets-to-your-github-repo>.
Alternatively, we can work on disabling signing or disabling the android build if the secrets are missing.

You should disable meta's automatic version code increment if you're using android builds that you build yourself. This appears to be off by default? Could be wrong, though.
 
Punted:
 - how do we want to manage versions?
   - as-is, it currently assumes that tags are versions, but does not require any tags to be present, nor prescribes any format for the tags.
 - how do we want to manage the android version code?
   - should this change on every commit, or only change when a release is made?
   - as-is, i'm just holding a constant androidVersionCode that is larger than the pre-existing ones to make downgrading easy.

Future Work:
 - [Automatic Deployments to Steam](https://game.ci/docs/github/deployment/steam)
   - Looks like default branch would still have to be managed manually.
 - [Automatically update a Personal License for CI](https://github.com/game-ci/unity-license-activate)
   - Not sure what kind of license is being used for the main repo, but may be useful for contributors setting up CI in their own repo.
 - Upload release artifacts to a draft release, then publish at the end
   - i.e. the idea recommended by https://github.com/marketplace/actions/publish-release
   - Could also automate mentioning the build in discord once it's finished building?
   - Currently, I've just been making a non-draft release in the UI. Mark as pre-release, then unmark as pre-release once the CI build finishes.

### DIFF
--- a/.github/workflows/compilation.yml
+++ b/.github/workflows/compilation.yml
@@ -22,22 +22,29 @@ jobs:
         if: "${{ env.TMP_SECRET1 != '' && env.TMP_SECRET2 != '' && env.TMP_SECRET3 != '' }}"
         run: echo "defined=true" >> $GITHUB_OUTPUT
 
-  build-client-win32:
-    name: Build client on Windows
+  build-client:
+    name: Build client for ${{ matrix.targetPlatform }}
     timeout-minutes: 100
-    runs-on: windows-2022
+    runs-on: ${{ matrix.buildPlatform }}
+    env:
+      projectPath: Basis
+      buildName: Basis Unity
     strategy:
       fail-fast: false
       matrix:
-        projectPath:
-          - Basis
-        targetPlatform:
-          - StandaloneWindows64
+        #targetPlatform:
+        #  - StandaloneLinux64
+        #  - Android
+        #buildPlatform:
+        #  - ubuntu-latest
+        include:
+          - buildPlatform: windows-2022
+            targetPlatform: StandaloneWindows64
     needs: [check-secret]
     if: needs.check-secret.outputs.secret-is-set == 'true'
     steps:
       - name: "Checkout repository"
-        timeout-minutes: 2
+        timeout-minutes: 5
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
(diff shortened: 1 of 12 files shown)


kitlith:
so, something I've noticed is that the Library cache is kinda Large. 2.4 GB for Windows, 3.5 GB for Android, and I'm currently waiting an hour to discover what it looks like for Linux on a fresh cache. (was 4.6 GB on the last run, but that was starting from the android cache.)

Github has a 10GB limit on cache, so even just keeping a single cache around for each of these platforms brings us to about the limit of what we can cache per repository. (it might be more efficient to somehow combine the caches from each platform into a single cache, given that linux cache was only 0.9 GB larger than the android cache when it was based off of it, but that would likely imply doing each build in sequence rather than in parallel. maybe group by what platform we're building on? so windows, android + linux caches?)

We may want to drop some platforms from CI. But... I don't think we really get any benefits from dropping platforms from CI unless we get the total cache between all platforms under 5GB, allowing us to have two copies of the full cache. (read: if we wanted to drop platforms to have a "useful" amount of cache, we might be stuck at 

there's also a hierarchy of which branches can use which caches. when building a branch, you can use caches from parent branches, but you cannot use caches from child branches.

Given all this information, we might be best served by choosing a branch (i.e. turning `doolys-branch` into `dev`, marking that as the main branch, and *only caching builds made on that branch*. (i.e. choosing to restore cache on all builds, but only updating the cache on commits to the dev branch.) Documentation referring to the lts branch would be updated to use `git clone -b lts <repo>`, etc.

This seems like a decent balance between having a useful cache, and not accidentally thrashing it out of existence given the imposed limits.

kitlith:
Alright. This should be ready for review.

Relevant example artifacts:
 - example release: https://github.com/kitlith/Basis/releases/tag/test
 - workflow run for release: https://github.com/kitlith/Basis/actions/runs/16361122079
 - workflow run for commit: https://github.com/kitlith/Basis/actions/runs/16361106949
 
 Cache management strategy is as follows:
  - Always attempt to restore cache, but only from the current platform.
  - Save cache whether the build succeeds or not, but only on the developer branch.
  - Automatically delete any cache entries that are not the latest for this platform.

This should keep our cache under 10GB and keep Github from deleting cache based on when it was last accessed rather than when it was created. Unfortunately, it'll likely still grow over time and have to be deleted, resulting in an hour long CI run at that time. Unfortunately, I'm not aware of any way to tell unity to remove entries from the Library folder that are not currently being used. If that *did* exist, I don't think we'd have to worry as much about the cache growing out of control.

I've added a couple more items as future work (should maybe turn them into issues?), but otherwise I think this is good to go, and I'd like to see how well it works in practice.

dooly123:
is there documentation on how to add secrets?
after that I will merge

kitlith:
@dooly123 where would you like the secrets documentation to go?

dooly123:
first time needing that kinda thing :) we could put it in one of the folders in the repo?

kitlith:
I'm thinking either a `CONTRIBUTING.md`/`HACKING.md` in the root of the repo, or a "Contributing" page on the wiki?

dooly123:
if you want, you can link it from the readme and make it its own file?

PR #308, opened 2025-07-23
Support base64-encoded Avatar password in URL Fragment

kitlith:
This feels a bit like spaghetti logic.

Originally, I was going to support query string syntax. i.e. `#password=blah&idk=<...>`, url decode, etc. However, using `System.Web.HttpUtil` requires messing with csc.rsp, which I wasn't a fan of. I'd like to hear opinions on if there's any other flags that might be nice to bundle into the URL (aka: do we need future extensibility?)

A possible alternative here is to register a URL scheme, and pass a URL + password in that scheme. (i.e. `basisvr://avatar?url=<...>&password=<...>`) which would allow someone to simply click on a link to add an avatar, assuming the url scheme has been registered.


yewnyx:
it's probably worth noting explicitly in the comment that the processed URL is overwritten to avoid the password improperly showing up in the servers' URL request logs

PR #312, opened 2025-07-30
Automatic docker builds for server

mriise:
This is roughly what it looks like https://github.com/mriise/Basis/pkgs/container/basis-server/473972740?tag=nightly

builds on developer (nightly) & long-term-support (latest)

please squash commit to hide my shame.

### DIFF
--- a/.github/workflows/build-docker-server.yml
+++ b/.github/workflows/build-docker-server.yml
@@ -1,0 +1,71 @@
+name: Build and Push Docker Server Image
+
+on:
+  push:
+    branches:
+      - developer
+      - long-term-support
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: basis-server
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


kitlith:
heads up, i'm not sure how effective the caching here is going to be, given how much we already use. <https://github.com/BasisVR/Basis/actions/caches>

though that particular issue would go away if the server was in a separate repository. i'm guessing that currently isn't done, to keep client and server network code in sync.

mriise:
Will see in a commit or two. I wanted to use the container registry to cache layers but github was showing the cache container as the latest build on the homepage while i was testing which i considered too confusing for people. This repo isnt showing container builds on the homepage so im inclined to switch back to that since it wont be using gha cache. 

~~If thats not gonna work we dont need to cache, its really not all that bad- especially since we build minutes are free :^)~~
Actually with how little we touch server code, the improvement in build time is kinda worth it anyway. Could probably try doing a diff to see if server code changed at all before running but that adds complexity that i dont want to bother with.

PR #317, opened 2025-08-06
Set OpenXR tracking origin to local_floor

SpookySkeletons:
As of yet untested, this may enable playspace movers to function in elevating the player off the floor while the client is on the OpenXR API.


dooly123:
no good sadly

PR #343, opened 2025-09-26
Migrate face tracking to newest Basis networking

hai-vr:
Migrate face tracking to newest Basis networking:
- Migrate to newer Basis avatar networking systems:
    - Remove GUID-based networking identifiers, which was a remnant of the previous networking system where BasisAvatarMonoBehaviour did not exist and all packets had to be routed manually to their destination components.
    - Migrate networking to the new BasisAvatarMonoBehaviour networking system, which handles packet routing on its own by transmitting packets based on the component rather than based on the avatar.
- Add Automatic Face Tracking component:
    - The Automatic Face Tracking component is now the entry point to adding HVR face tracking capabilities to the avatar.
    - This component automatically detects all SkinnedMeshRenderer components on the avatar that have ARKit or Unified Expressions blendshapes on it.
    - By default, this component uses the basic face tracking definition files that are bundled with the application, but this can be either overridden by the user or supplemented.
        - Due to this, the face tracking assets are now addressables.
    - This components automatically starts an OSC service on the wearer.
- Improve the handling of face tracking data:
    - Add Address Overrides to the actuation definition file, so that we can define default values for the eyelids.
    - Work towards the mutualization of the interpolation of addresses between Eye Tracking and Face Tracking.
    - Actuators now introspect which blendshapes actually exist on the target renderers, and will no longer try to network any actuator that has no blendshape associated with it.
    - Add more debugging in the blendshape actuation definition file editor.
    - Fix TongueRoll actuator was targeting the wrong blendshape.

Change VRCFaceTracking setup process:
- In the Automatic Face Tracking component, the user now presses a button to create the face tracking JSON file used by VRCFaceTracking.
- We want to ensure we overwrite any previous version of face tracking that may have already been installed by previous users of the previous system, so we're changing the dummy avatar ID to a different one. That way, we're sure this is really the file that will be loaded by VRCFaceTracking.

Move the package to dev.hai-vr.basis.comms/

Add MIT License file to dev.hai-vr.basis.comms/ to make it explicit the package itself is also MIT licensed.


Basis/Packages/dev.hai-vr.basis.comms/Scripts/Systems/Runtime/Components/FaceTracking/AutomaticFaceTracking.cs
```diff
@@ -0,0 +1,245 @@
...
+            overrideDefinitionFiles = HVRCommsUtil.SlowSanitizeEndUserProvidedObjectArray(overrideDefinitionFiles);
+            supplementalDefinitionFiles = HVRCommsUtil.SlowSanitizeEndUserProvidedObjectArray(supplementalDefinitionFiles);
+
+            if (_avatar == null)
+            {
+                _avatar = HVRCommsUtil.GetAvatar(this);
+            }
+
+            _ueHandle ??= Addressables.LoadAssetAsync<BlendshapeActuationDefinitionFile>("HVR.Basis.Comms.FaceTracking.DefaultUnifiedExpressionsDefinitionFile").WaitForCompletion();
+            _arKitHandle ??= Addressables.LoadAssetAsync<BlendshapeActuationDefinitionFile>("HVR.Basis.Comms.FaceTracking.DefaultARKitDefinitionFile").WaitForCompletion();
```
hai-vr, line 51: I need your attention on this because the default face tracking definition files need to be part of the application build process (and not just the avatar). These have been added to `Basis Foundation Assets.asset`, but is there anything else that needs to be done?

dooly123: You’re also welcome to reference them directly from a field. That way you don’t need to worry about it, though it will duplicate memory. Still, that’s fine—let’s go with this approach.

hai-vr: The point of using addressables here is that by default, if the definition file updates at the package level, then the users will have the most up to date definition file without needing to update the avatar.
It's because I'm now treating the face tracking definition files are being part of the application rather than user-generated content; with the user still being able to override the application through the component settings if they want to.

dooly123: understandable

hai-vr:
Just gonna write down my findings regarding an issue that is preventing this PR from getting out of draft:

I tried this on my machine, and I use a very old version of VRCFaceTracking compiled from source, which works.

However, as part of a test, I've tried to run this my other machine which uses the Steam version of VRCFaceTracking, but the Steam version of VRChatFaceTracking refuses to detect an avatar change (and subsequently, will not try to load the JSON file). The beta branch of VRCFaceTracking on Steam doesn't detect the change either.

In order to debug why that's happening, I've compiled the latest version of VRCFaceTracking from the source code from GitHub on my other machine. However, that version, compiled from source, works. It can detect the avatar change and load the JSON file.

So far I don't know what's happening. Merging this PR does not make sense if users won't even be able to use it, so this needs further investigation.

PR #345, opened 2025-09-28
Fix remote should not send messages to all users

hai-vr:
- When the avatar spawns on remote, the initialization packet should only be sent to the avatar wearer, not all the users.
- This should fix "Illegal sender" error.
- Force the initialization packet to be reliable.


hai-vr:
Opened by request

PR #349, opened 2025-10-01
Auto-rename duplicated humanoid bone names

hai-vr:
- The AnimationRigging package currently has an issue: If a humanoid bone transform shares the same name as a non-humanoid bone transform inside the avatar, it will refuse to run with an error (the TransformStreamHandle cannot be resolved).
- This is an issue for avatar users, as it is extremely commonplace for bones to share the same name as a humanoid bone when merging costumes into an avatar base.
- This commit changes the behaviour when building or testing an avatar: By default, we will now detect any non-humanoid bone that shares the same name as a humanoid bone, and rename those non-humanoid bones by appending _1, _2, to its name, and so on.
- This default behaviour can be opt-out of by going to Basis Avatar component > Processing Options > Check "Do not auto-rename bones".
- The SDK will only show warnings for duplicate bone names as long as this option remains unchecked.
- A new serialized data structure has been added to Basis Avatar: BasisProcessingAvatarOptions.
- As this data structure has no use at runtime, this is set to null on the built (cloned) avatar when it finishes processing it.


dooly123:
looks good

PR #351, opened 2025-10-02
Fix Test in Editor should not use simulated Jiggles transforms

hai-vr:
- When entering Test in Editor, the copy of the avatar that is being instantiated was using the Jiggle transforms post-simulation.
- As a workaround, if Jiggles are detected on the avatar when pressing Test in Editor, turn the avatar off for three frames before instantiating a copy of it.


dooly123:
i have pushed up the fixes from naelstrof let me know if you still need this merged ty.

hai-vr:
This can neither be merged nor closed because as far as I'm aware the newer version of JigglePhysics does not seem to completely fix the issue that it intended to fix.

Troubleshooting described in Discord:
https://discord.com/channels/1239242259392757822/1288184592930570293/1423412210541269028

image

To make it work, I have to:
- Explicitly call SnapToRestPose on the JiggleRig that were enabled post-Instantiate, and
- Disable the `t == rootBone` check

The source code of JigglePhysics specifically ignores rootBone in the SnapToRestPose process and I'm not knowledgeable with the domain to have the capacity to understand its justification

image

dooly123:
im going to  merge this for now and correct once nael has a better fix

PR #366, opened 2025-11-02
Add server images for all linux, and docker images

Toys0125:
changed the os selection to be runtimes to add docker and arm support.


Toys0125:
I feel like an idiot for not noticing that last matrix...

PR #368, opened 2025-11-05
Pickup enhancements

foxriot:
Adds three optional abilities to the interaction system:

- "Scale With Gesture" which smoothly scales an object held in one hand while moving the other hand  (hand-trackers only)

- "Lock To Axis": Axis-based constraints (x,y,z or none) with clamping limits

- "Auto Return" which transforms a manipulated object back to its original scale, position and rotation after being left alone for a user-defined duration of time. Supports standard and custom easing curves. 

image


foxriot:
Opening a cleaner one in a moment

PR #369, opened 2025-11-06
Pickup Enhancements

foxriot:
Adds three optional abilities to the interaction system:

"Scale With Gesture" which smoothly scales an object held in one hand while moving the other hand (hand-trackers only)

"Lock To Axis": Axis-based constraints (x,y,z or none) with clamping limits

"Auto Return" which transforms a manipulated object back to its original scale, position and rotation after being left alone for a user-defined duration of time. Supports standard and custom easing curves.

509898139-cb894989-38cc-409d-8bde-aca3eb315a86


foxriot:
I feel like I may have missed some of the improvements we covered - feel free to ask me to review anything again and/or just make changes, I know you wanted to try and get it in this week so I hustled a bit.

dooly123:
thankyou

PR #376, opened 2025-11-16
Experimental Persistent KV store

mriise:
This adds a generic persistent kv interface for server side values. The intent is this will eventually be exposed to clients.

The current backing database is SQLite.

DATABASE STRUCTURE UNSTABLE! You may lose data until it is considered stable.

- There is no networking (clients cant use it in any way atm)
- There is no way to configure limits for buckets/keys or database settings
- Performance is not measured
- Tests are AI generated (take with a grain of salt)

~~@dooly123 DO NOT MERGE DRAFT!~~


TheButlah (commented):
Seems very useful :3
how difficult would it be to get the KV store to work on either client or server rather than only supporting server KVs? One could imagine keys being labeled as persisted either locally (and are client-authoritative) or on the server (and are server-authoritative).

mriise:
we would need: 
a) strong embedded db (sqlite preferably)
b) stable format for that db

in reality you dont need a sync mechanism just to _have_ a client kv store (we can do it independently), but for what you are asking we would.

In many cases though i would think that you just want to keep things in-memory, then persist on exit of the app or instance. That could just as easily be done with file serialization.
this last point i would defer to @dooly123 on however.

TheButlah:
> you just want to keep things in-memory, then persist on exit of the app or instance

That same line of reasoning could be applied to indicate that we should not use sqlite on the server, and instead just persist things on exit as json. But in reality, I think its beneficial to have something flushed to disk when the function exits, rather than when the program exists, and sqlite is a known and reliable way to transactionally do that.

The same rationale that justifies its use on the server, justifies its use on the client (at least, imo)

mriise:
fair enough, though doing so is out of my scope for the forseeable future 

thankfully nobody really uses the existing database so i dont think we will need to adapt that either if dooly considers this to be core enough.

doing migrations for clients too is a bit of a yike, so id like to settle on a db structure that is stable first.

yewnyx:
How would it work in the local case where someone is hosting a session from their client?

mriise:
versioned and time-stamped diff sync or something of the like, basically treat the server as a cache/broadcast layer. but again, you dont need server sync to add client kv.

mriise (commented):
Tests are a bit verbose, but fine enough. Some of them reference the static class, so that will need updating when another backend is added (and a management interface is settled)

mriise:
@dooly123 ready for review/merge

PR #388, opened 2025-11-22
Refactor local player height code

aaronfranke:
The existing local player height code was a bit of a mess. This PR is my attempt at refactoring of it, though a few issues remain.

Benefits:
- Unified `BasisLocalHeightInformation` and `BasisHeightDriver` into `BasisLocalHeight`.
- The properties of `BasisLocalHeight` are read-only for public users, so setting these values is now encapsulated.
- Much clearer naming and separation of avatar vs player scale terms. Replace "default" with "fallback" for clarity.
- Introduce "RealUser" named properties as a way to properly refer to the real user's sizes.
  - These currently aren't read from elsewhere in the codebase, which is a problem.
- Avatar heights/scales are now saved and loaded per-avatar, allowing different avatars to have different scales.
- New ability in the code to set the height as a scale multiplier instead of an amount of meters via `BasisHeightMeasurement.ScaleMultiplier`. Could be exposed in the UI later.
- New functions in `BasisDataStore` to save and load JSON in a well-formatted cross-platform way.
- The near clipping plane will automatically pull itself closer inwards when using a small avatar, which tries to avoid the avatar becoming invisible or the menu becoming unusable at small avatar sizes.

Broken things:
- In VR, the view scale/position vs the ground doesn't change. So when you resize the avatar to be larger, more of the avatar goes into the ground, and when you resize the avatar to be smaller, the legs float in the air.
  - I tried to solve this issue for a few hours but I really don't know where the problem is or how this code is supposed to work, so I'll leave it to @dooly123 to fix it. For this PR I've kept the changes minimal.
  - The fix probably has something to do with using the new "RealUser" properties in some places, but I'm not sure where.
  - Like, the old code having some level of confusion between the terms makes it difficult to truly figure out what the *intention* of the old code was.
- Some of the old code for calibration seemingly relied on changing the player's scale, which is... strange. I think, ideally, calibration should work locally together with whatever in-game scale the player is at, such that the calibration process only calibrates the real-world scale. This code is probably broken and needs fixing.

However, desktop mode works perfectly as far as I can tell.


dooly123:
Heyo, I have looked at this code a bit now.

The confusing structure and names have warped your understanding of the height system


The height system is broken up into multiple parts

1. scaling your devices to match the avatar you're wearing

2. fake scaling on the y to allow y height adjustment by the seated / standing mode

3. matching your IPD to stop rendering issues and distancing mistakes (scaling the camera)

4. The Avatar Scale is poorly named in the menu and will be getting updated with the newer menu. Its actual intent is in meters, how large your avatar can be overwritten to be up to 5 meters. This will be going into our avatar hand menu at some point. I like the idea of persistent scale offsets but we gotta do the math in order and not just apply the data.

dooly123:
i do like the idea behind the near clip plane ill give it some thought

dooly123:
i have also made some of the other name changes and will continue todo so for a bit.

dooly123:
closed this one so we can break it up further

PR #397, opened 2025-11-24
[DISCUSS] Convert asmdef's from GUIDs to names to faciliatate fragile situations

cnlohr:
…s where GUIDs may be lost.


dooly123:
Based on how Unity's own teams don't use GUIDs but use names, we will follow the same approach. It's really dumb, but that's just the way it is.

PR #399, opened 2025-11-24
Syncing differences between Basis Server copies

kitlith:
PR-ing this first, in case I got any of the changes wrong.

I *really* want to get rid of the duplication here. Having the Basis unity project reference the top-level folder for the Basis Server doesn't work because of user confusion. Using symlinks doesn't necessarily work because of Windows.

Can we could get rid of the top-level Basis Server folder, and just have the copy that's inside of the unity project? Or make the top level Basis Sever csprojs reference the copy that's inside of the unity project?


kitlith:
Never mind, you did it in https://github.com/BasisVR/Basis/commit/df0b7a74e84fb2b60aa7884f8d565edb7be44d3e

PR #407, opened 2025-11-28
Move height changed signal/action into BasisLocalHeight

aaronfranke:
The idea here is to centralize height-related code into `BasisLocalHeight` rather than having it be spread around. This is not the last PR organizing height-related code, but it is an easy first step.

### DIFF
--- a/Basis/Packages/com.basis.framework/Players/Local/BasisLocalHeight.cs
+++ b/Basis/Packages/com.basis.framework/Players/Local/BasisLocalHeight.cs
@@ -27,6 +27,11 @@ namespace Basis.Scripts.BasisSdk.Players
         public string AvatarName;
 
         /// <summary>
+        /// Fired on the frame after a player height change is requested.
+        /// </summary>
+        public static System.Action OnHeightChangedNextFrame;
+
+        /// <summary>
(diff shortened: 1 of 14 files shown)


Basis/Packages/com.basis.framework/Players/Local/BasisLocalHeight.cs
```diff
@@ -26,6 +26,11 @@ public class BasisLocalHeight
         /// </summary>
         public string AvatarName;
 
+        /// <summary>
+        /// Fired on the frame after a player height change is requested.
+        /// </summary>
+        public static System.Action OnChangedNextFrame;
```
dooly123: OnChangedNextFrame should really still be OnHeightChangedNextFrame

what changed the entire thing? thats not right.

aaronfranke: It's already in a class named `BasisLocalHeight` though, so every line of code using this already has "height" in the line.

EDIT: Oh, I see, it can be confused with "the entire BasisLocalHeight class's data changed".

dooly123:
I'm having a hard time with this one mentally, as I don't see the reason to change it, to me it's an action callback on the player's height changing, not the height on its own

PR #420, opened 2025-12-04
Allow multiple colliders per interactable and allow raycasts inside them

aaronfranke:
While doing more testing on the seat system, I wanted to allow 2 different things:

1. Placing the collider on a child GameObject, to potentially allow for multiple colliders per seat and other interactables.
2. Raycast hitting from inside of a trigger volume, so that you don't get stuck if you sit in a seat set up like that.

This PR updates the code to handle both of these things.
- Unify the collider referencing logic for BasisInteractableObject.
  - It now has the following rules:
    - If set to a non-empty array, these colliders will be used as the interactable's colliders.
    - If empty or null, and a collider exists on the same GameObject, that collider will be used.
    - If empty or null, and no collider exists on the same GameObject, all child colliders will be used.
  - Caching now happens for all BasisInteractableObject classes, avoiding getting the components every frame.
  - The desktop extra reach system now applies to all BasisInteractableObject classes.
  - New function `GetClosestPoint` that is split out of `IsWithinRange`.
- In `BasisPlayerInteract`, check parent GameObjects for the BasisInteractableObject component, but only consider the first found one, and only consider if that BasisInteractableObject's colliders include the hit collider.
- Update `BasisPointRaycaster` to cast backwards to enable hitting objects you're inside of. It has a size of 4 which allows minimally checking for being inside multiple colliders at once in case there are overlapping trigger volumes.
  - Note that this will not work for triggers whose boundary is farther than the raycast distance (30 meters), but let's be real who is making trigger colliders 30 meters across?
- Update `SeatScene.unity` to include a new seat on the west side of the platform which tests both of these cases at once.

Screenshot 2025-12-04 030352


Basis/Packages/com.basis.framework/Interactions/BasisInteractableObject.cs
```diff
@@ -251,30 +260,62 @@ private void OnInputRemoved(BasisInput input)
...
+            return Vector3.Distance(GetClosestPoint(source), source) <= interactRange + extraReach;
+        }
+
+        public Vector3 GetClosestPoint(Vector3 source)
         {
-            Collider collider = GetCollider();
-            if (collider != null)
+            float closestDistanceSqr = float.MaxValue;
+            Vector3 closestPoint = transform.position;
+            foreach (Collider childCollider in _colliderRefs)
```
mriise: nit: I would avoid foreach like the plague in unity if its per-frame. It will sometime do a temporary box for the IEnumerator interface. It shouldn't here however.

Basis/Packages/com.basis.framework/Interactions/BasisInteractableObject.cs
```diff
@@ -251,30 +260,62 @@ private void OnInputRemoved(BasisInput input)
...
-            return Vector3.Distance(transform.position, source) <= InteractRange;
+            return closestPoint;
         }
 
         /// <summary>
         /// Gets the collider attached to this object if one exists.
         /// Override with cached reference when possible.
         /// </summary>
-        public virtual Collider GetCollider()
+        public virtual Collider[] GetColliders(bool cache = true)
```
mriise: nit
```suggestion
        public virtual Collider[] GetColliders(bool useCache = true)
```

mriise: If GetColliders can change colliders at runtime, then we need to also recalculate the highlight clone(s) at runtime. Doing so is a fairly expensive operation (Object instantiation and mesh generation) so I would recommend having a different method entirely to change the colliders.

Basis/Packages/com.basis.framework/Interactions/BasisPlayerInteract.cs
```diff
@@ -157,10 +157,28 @@ private void PollSystem()
...
+                        if (hitInteractable != null)
+                        {
+                            Collider[] colliders = hitInteractable.GetColliders();
+                            if (colliders != null && colliders.Length > 0 && colliders.Contains(rayHit.collider))
+                            {
+                                isValidRayHit = true;
+                            }
+                        }
+                    }
+                }
```
mriise: This is really deep for if statements, if you can put this in a seperate function or otherwise simplify this do so. Player interact is already a hairy mess that needs less not more of this.

Basis/Packages/com.basis.framework/UI/BasisPointRaycaster.cs
```diff
@@ -122,6 +122,25 @@ public void UpdateRaycast()
                     ClosestRayCastHit = new RaycastHit();
                 }
             }
+            // One last thing: Cast backwards just in case the origin of the ray was inside a collider.
+            {
+                float backcastDistance = ClosestRayCastHit.distance > 0 ? ClosestRayCastHit.distance : MaxDistance;
+                Ray backcastRay = new Ray(ray.origin + ray.direction * backcastDistance, -ray.direction);
+                RaycastHit[] backcastHits = new RaycastHit[4];
```
mriise: This is allocating a new RaycastHit array every frame. keep the scratch buffer here in a variable on the class.

mriise (changes requested):
One major thing (what to do about highlight clones), a few small things, and a few nits.

Basis/Packages/com.basis.framework/UI/BasisPointRaycaster.cs
aaronfranke: Dang, is C# not smart enough to avoid the heap for local temporary arrays?

Basis/Packages/com.basis.framework/Interactions/BasisInteractableObject.cs
aaronfranke: I'm going to remove this function's ability to change the colliders at runtime.

Basis/Packages/com.basis.framework/UI/BasisPointRaycaster.cs
mriise: that would require a smart JIT, or a really smart compiler of which il2cpp/mono is not.

mriise (approved):

PR #459, opened 2025-12-17
Change tracked role assignment in BasisInput.cs (Fixes Vive Controllers)

GapVR:
Fixes the "with bone LeftHand as HasTracker" causing all Vive Controllers to get assigned to LeftHand.

Controllers need to be powered on and active in SteamVR before switching to OpenVR mode in Basis.

Caveat: I have no idea what this will do to other controller types.


dooly123:
This is not a fix. Just change the devices scriptable object

dooly123:
i have pushed a fix that should resolve this issue at its root

PR #460, opened 2025-12-17
Add Zeromessenger

mriise:
I will be using this library to do message processing for avatar state. It is not terribly complex and is already designed for use in unity. I may adapt this to support batching messages in the future (and process in jobs) but I'll do so when profiler tells me i should.

Avatar state is a very open topic and there will be may different ways to modify local and process remote state. By using this library we can experiment with how inputs, osc, ect affects the avatar without needing to refactor (and potentially have it runtime plug able)

It is a near exact fork of https://github.com/annulusgames/ZeroMessenger


yewnyx:
nit: "messanger" is (i presume) a misspelling

mriise:
correct

PR #462, opened 2025-12-19
LTS update 19 12 2025

dooly123:
The last LTS release was before we attempted 600+ users. Since then, we have made further improvements and fixed bugs 

The differences are bug fixes, ui changes, and IK changes


aicodeguard:
### Code Review – Key Findings

**Summary:**
This review identified 3 critical issues, primarily focusing on security vulnerabilities where sensitive credentials (passwords) are logged in plain text, and a concurrency bug that could lead to system crashes.

## Issues

1. **[Security] Plain-text Credential Logging in Client Console**
   The `ReadString` helper method logs every configuration value it reads, including `Password` and `AvatarPassword`. This exposes credentials in the application logs.
   - **File:** `Basis Server/BasisNetworkClientConsole/BasisNetworkClientConsole/ConfigManager.cs`
   - **Line:** [30](https://github.com/BasisVR/Basis/blob/432e3d8b082f7b687efd9c80fab0fe9509f4d364/Basis%20Server/BasisNetworkClientConsole/BasisNetworkClientConsole/ConfigManager.cs#L30)
   - **Fix:** Remove the `BNL.Log` call in `ReadString` or filter out keys containing "Password".

2. **[Security] Plain-text Server Password Logging**
   The `ProcessEnvironmentalOverrides` method iterates through all configuration fields and logs their values if an environment variable override exists. This includes the server `Password` field.
   - **File:** `Basis Server/BasisNetworkCore/BasisServerConfiguration.cs`
   - **Line:** [100](https://github.com/BasisVR/Basis/blob/432e3d8b082f7b687efd9c80fab0fe9509f4d364/Basis%20Server/BasisNetworkCore/BasisServerConfiguration.cs#L100)
   - **Fix:** Modify the logging logic to check if the field name is sensitive (e.g., "Password") and redact the value in the log output.

3. **[Bug] Race Condition in Static Initialization**
   The `AvatarSyncSize` property uses a non-thread-safe lazy initialization pattern. `IsWired` is set to `true` before `avatarSyncSize` is calculated. A concurrent thread could read `IsWired` as true and return `avatarSyncSize` as 0, causing downstream buffer allocation failures.
   - **File:** `Basis Server/BasisNetworkCore/Compression/BasisBitPackingConstants.cs`
   - **Lines:** [17-22](https://github.com/BasisVR/Basis/blob/432e3d8b082f7b687efd9c80fab0fe9509f4d364/Basis%20Server/BasisNetworkCore/Compression/BasisBitPackingConstants.cs#L17-L22)
   - **Fix:** Move the initialization to a static constructor or use a `lock` block to ensure atomicity.

yewnyx:
```
                if (IsWired == false)
                {
                    IsWired = true;
                    //156 = 12 + 8 + 2 + 134
                    avatarSyncSize = WritePosition + WriteRotation + WriteScale + SumBitsPerSlot();
                }
                return avatarSyncSize;
```

Just FYI, this pattern of lazy initialization is definitely not threadsafe, if you use it anywhere else.

Also, why does your constants file hold public non-constant non-readonly integer variables? Also the byte arrays are mutable (their reference is readonly, but not their contents).

Toys0125:
> Just FYI, this pattern of lazy initialization is definitely not threadsafe, if you use it anywhere else.
> 
I find it interesting that the aicodeguard says the exact same thing. Only 20 mins before you.

yewnyx:
> > Just FYI, this pattern of lazy initialization is definitely not threadsafe, if you use it anywhere else.
> 
> I find it interesting that the aicodeguard says the exact same thing. Only 20 mins before you.

Not at all. I took a look at that to see if it was actually worthwhile and think it is worth a human actually saying "yes this is real"

PR #471, opened 2025-12-22
Change the KeyStore location to My Documents on Desktop and Unity Editor.

Toys0125:
Making this change because the currently implemented location for the keystore is not protected by most backup programs. Too many files and cache files are stored in AppData/LocalLow. So wanted the keystore to be protected by moving it to my documents.

This change only affects desktop which majority of distros will have $HOME set and still follows the normal, Company Name, Products name scheme.

Added Simple Backup if the file exists just copy it to a bak extension.


dooly123:
when i get time i will make this a selectable path, changing it now doesnt make sense but will soon

PR #484, opened 2025-12-31
More robust validation of avatar texture settings

BoatFloater:


Toys0125:
In my opinion, needs to be a more advanced check due to some people will forget to enable mipmaping for their avatar textures. And if you enable the streaming even though mipmap is off will pass the check currently and wont affect the texture since its off. So might force enable streaming textures for all textures will be better.

BoatFloater:
Mip maps are enabled by default; if they are disabled that means the person who imported them specifically turned them off.
The check is not relevant to textures with disabled mip maps. I think it's better for this check to be accurate than to use it as a nudge to encourage users to enable mip maps on various textures.
Some textures, for example shader LUTs, might come with mip maps off and could confuse the user why they are getting a warning about it.

dooly123:
i think we need todo more to identify ignorable textures instead of just not warning, luts tend to be small and toys made a good point that really small textures could just be skipped to solve most of the issues at hand

BoatFloater:
Sure, but I don't understand why there should be a warning instructing the user to enable mip streaming on a texture where 'Generate Mipmap' is not checked
image

dooly123:
mipmap streaming is one of the only things we have to reduce the memory footprint of avatars / uploaded content.
It being off is frankly painful for anyone to deal with. It should be on unless it's explicitly an issue. 
Bypassing the rules is something people do all the time, and I don't like that. Nor should anyone else.

BoatFloater:
But the check does not validate or enforce 'Generate Mipmap'. Only 'Stream Mipmap Levels', which does nothing if 'Generate Mipmap' is off.

dooly123:
ok i see, yes that could be improved

dooly123:
The fix you have is a Band-Aid and will lead to missed opportunities for performance gains from streaming.

we need to improve this further.



1. detect what kind of texture mode we are looking at.

2. use some assumptions that make sense (size is below 256 / 128)

3. depending on those outcomes display different warnings  / no warnings

BoatFloater:
I made the checks more robust:
* check for 'Generate Mipmap' on textures >256
* check for 'Stream Mipmap Levels' on all textures which have 'Generate Mipmap' enabled
* Added Auto-Fix actions for both checks and the existing check for >4k textures

I also noticed some issues with the auto-fix buttons (they were gray before the first hover and the text overflowed) so I fixed those and added a warning color and optional action title; let me know if that would be better resolved another way.

dooly123:
Appreciated

PR #487, opened 2026-01-02
iOS audio crash and speaker default output

katruud:
iOS clients to can connect to compatible servers and can interact with other users
Windows/Mac builds should have no changes but tested as well

Updated libopus to avoid symbol conflicts and crashes
Add XCFramework to combine libopus iOS libraries (device, simulator, ARM, x86)
Updates for iOS static library requirement
Default audio output to speaker instead of earpiece

todo:
demo world is too large for iOS phones and will crash
audio is quieter on iOS
simulator iPad build appears dark


katruud:
XCFramework/libopus for iOS compiled at https://github.com/katruud/OpusSharp (PR open for https://github.com/AvionBlock/OpusSharp)

PR #491, opened 2026-01-04
Avatar panel update (add mac and ios images)

towneh:
Edits the script to accept additional os platforms in the avatarlist prefab and introduces new icon images for the platforms


towneh:
PR contains changes which remove the text and leaves only the icons due to space contraints

Appearing for avatars built on all OS platforms:
image

Seems to be incorrectly counting Mac and iOS for LoadingAvatar:
image

PR #520, opened 2026-02-05
Update Cilbox & Whitelist Binary/Stream Classes

BoatFloater:
* Update Cilbox to latest master (required for the new classes)
* Whitelist the classes required to use `BinaryWriter`/`BinaryReader` to make network sync much easier in Cilboxable components


yewnyx:
pedantic nit, but:

> Whitelist Binary/Stream Components

Classes*

BoatFloater:
Oops.. I even wrote classes in the description/commit. Apparently I don't have permission to edit the PR title

dooly123:
im happy to merge this but i just need a few more eye balls to atleast see if there would be a security risk

dooly123:
looked into each class, should be ok as long as we never expose file handling  / expose it in a way were this can effect

PR #537, opened 2026-02-16
use animator transform for PoseHandler in remote players

BoatFloater:
My avatar's animator is a child of the `BasisAvatar` component, which does not seem like it should be unsupported according to the docs and some comments I found in the code.

I changed the PoseHandler to reference the animator's transform (instead of the root transform of BasisAvatar) which seems to align with the [code on the sender side](https://github.com/BasisVR/Basis/blob/6795bd3eadfe4bdedfe57398329571e58ffd2493/Basis/Packages/com.basis.framework/Networking/NetworkedAvatar/BasisNetworkAvatarCompressor.cs#L51-L52).

I only tested this fix in desktop mode but before the change, the avatar was stuck in T-Pose at origin on remote clients, and after the change, it seemed to work normally.


dooly123:
That will cause a performance degradation. I'm happy to merge it if you pre-store the transform as the other one was.

dooly123:
crossing my fingers that its all ok :)

basically desktop and vr are identical in operation over the network so should all be good

PR #547, opened 2026-02-26
Update Cilbox

cnlohr:


BoatFloater:
@cnlohr `CilboxSceneBasis` shim needs the `CheckFieldAllowed` override before this is merged I think

cnlohr:
Ah! Good point.  Will poke at it tomorrow.

BoatFloater (approved):

PR #553, opened 2026-02-27
organize CilboxSceneBasis, add CilboxPropBasis, modify BasisLocalPlayer

BoatFloater:
A few cilbox related changes in this PR.

* CilboxSceneBasis
  * I sorted all the types allowed. This should make it easier to edit and review the whitelist (well, after this PR..)
  * The only actual change to the whitelist is BasisLocalPlayer with tight restrictions:
    * `GetPositionAndRotation`, `Teleport`, `Respawn`, `get_Instance`
* CilboxPropBasis
  * I saw that the content police for props currently reference the `CilboxScene` from the Cilbox package, but it's probably better to control the rules internally to the Basis project, so I created `CilboxPropBasis` with the exact rules from `CilboxScene` unchanged.
* PropContentPoliceSelector
  * Reference `CilboxPropBasis` instead of `CilboxScene`
* BasisLocalPlayer
  * Add GetPositionAndRotation
  * Change `Instance` from a field to a field accessor with private setter
    * Generally this is source-compatible but binary incompatible, so I can approach this another way if you'd like.

### DIFF
--- a/Basis/Packages/com.basis.framework/Players/Local/BasisLocalPlayer.cs
+++ b/Basis/Packages/com.basis.framework/Players/Local/BasisLocalPlayer.cs
@@ -30,7 +30,7 @@ namespace Basis.Scripts.BasisSdk.Players
         /// <summary>
         /// Singleton-like reference to the active local player instance.
         /// </summary>
-        public static BasisLocalPlayer Instance;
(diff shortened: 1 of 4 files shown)


dooly123:
thankyou for the time put into this

Issue #554, opened 2026-02-28
Can't Hold Pickup Objects and Press Trigger When in VR


Chylum:
### Describe the bug?

Players can't hold a BasisPickupInteractable and press the trigger button. The input event doesn't fire due to LateDoPolldata() is empty for both OpenVR and OpenXR hand inputs. This means the CurrentInputState never gets updated between the last copy and InputUpdate().

Desktop works because BasisDesktopEye's Latepolldata() copies a fresh mouse state over.

### Steps to Reproduce

Add a BasisPickupInteractable component to any GameObject designed to use the trigger button to do something (e.g., fire a weapon or draw something).

### The expected behavior

The trigger should do fire something!

### Build Info

Unity 6.3.5f1

### Log Files

<img width="710" height="785" alt="Image" src="https://github.com/user-attachments/assets/cd882a45-bd43-4538-8675-deae0513e723" />

<img width="991" height="780" alt="Image" src="https://github.com/user-attachments/assets/eab191a7-0c2d-41ee-84e0-8f4a1e17a7b0" />

<img width="991" height="206" alt="Image" src="https://github.com/user-attachments/assets/e4dc06a5-6efa-4472-8cdc-16cc91a9e1ce" />

### Additional Context

This is the fix, coming up:

<img width="960" height="202" alt="Image" src="https://github.com/user-attachments/assets/9bc8b74f-20f4-4c98-b0ea-8c7314f2167e" />

dooly123:
was merged

PR #555, opened 2026-02-28
Fix Trigger Not Working When Interacting with Pickup Objects in VR

Chylum:
#554


dooly123:
i hate the fact that we are reading it 2x but looking at this more since 2 years ago yer its making more sense ;/

PR #582, opened 2026-03-07
fix(face-tracking): make host authoritative for activity state and neutralize startup sync

Toys0125:
* Added an internal `FaceTrackingActivityRelay` channel (`HVR/Internal/FaceTrackingActive`) with a host-only inactivity timeout (0.5s). Remotes now follow host activity and do not run their own timeout logic.
* Wired OSC input to activity updates so any incoming face packet marks tracking active.
* Fixed remote eye application in `EyeTrackingBoneActuation` (!IsLocal && Receiver != null) and ensured remote override flags are cleared when tracking is inactive/disabled.
* Gated eye and blendshape actuation behind propagated tracking activity:
Active: HVR drives local/remote face data.
Inactive: reset HVR-driven values to neutral and return control to Basis default behavior.
* Stopped applying startup/default blendshape overrides when tracking is inactive; defaults are only pushed while tracking is active.
* Changed mutualized feature initialization to neutral normalized values based on interpolation ranges (AbsoluteToRange(0) when 0 is in range), preventing signed channels from booting at low-end values (e.g. unintended TongueLeft).
* Updated networking init flow to seed neutral values before initial replay, and fixed wearer/range bookkeeping issues in `HVRAvatarComms`.

### DIFF
--- a/Basis/Packages/com.basis.eventdriver/BasisEventDriver.cs
+++ b/Basis/Packages/com.basis.eventdriver/BasisEventDriver.cs
@@ -230,10 +230,12 @@ public class BasisEventDriver : MonoBehaviour
     /// </summary>
     private void OnBeforeRender()
     {
+        // Ensure remote face jobs are completed/applied even before local player is ready.
+        BasisRemoteFaceManagement.Apply(); //apply blendshapes
+
         if (BasisLocalPlayer.PlayerReady)
         {
             BasisLocalPlayer.Instance.SimulateOnRender();
-            BasisRemoteFaceManagement.Apply(); //apply blendshapes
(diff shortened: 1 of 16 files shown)


dooly123:
break out the other changes to just what is needed and run that by hai please

Toys0125:
> break out the other changes to just what is needed and run that by hai please

Okay done.
@hai-vr if you up to review the changes be greatly appreciated.

dooly123:
I have tested this a little bit. I'm going to merge it this week, so we can at least test it for a week and find out if it's better or worse

dooly123:
was indeed just better

PR #600, opened 2026-03-10
Add setting for microphone behavior at startup

BoatFloater:
The main feature of this PR is adding a setting to control what the microphone does at startup. The options are:
* Muted (new default)
* Unmuted (current behavior before merge)
* Remember Last State

I also hid the microphone icon until the microphone driver is initialized, so it doesn't show at the wrong color and position at start.

Let me know if you would like me to change the default or any of the verbiage.

### DIFF
--- a/Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
+++ b/Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
@@ -202,7 +202,7 @@ namespace Basis.Scripts.Drivers
             }
 
 #if !BASIS_DISABLE_MICROPHONE
-            microphoneIconDriver.SpriteRendererIcon.gameObject.SetActive(false);
+            microphoneIconDriver.HardEnableVisuals(false);
             BasisLocalMicrophoneDriver.OnInitializedAction += OnMicrophoneDriverInitialized;
 #endif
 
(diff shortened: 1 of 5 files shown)


Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -202,21 +202,15 @@ public void OnEnable()
             }
 
 #if !BASIS_DISABLE_MICROPHONE
-            microphoneIconDriver.Initalize(this);
-            microphoneIconDriver.UpdateMicrophoneVisuals(BasisLocalMicrophoneDriver.isPaused, false);
+            microphoneIconDriver.SpriteRendererIcon.gameObject.SetActive(false);
```
dooly123: This really should be a function on the microphone driver, so it's more easily reusable/understandable

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalMicrophoneDriver.cs
```diff
@@ -110,18 +114,35 @@ private static bool IsPaused
...
 
+    public static bool ResolvePausedFromSettings()
+    {
+        string behavior = Basis.BasisUI.BasisSettingsDefaults.MicStartBehavior.RawValue;
+        switch (behavior)
+        {
+            case SettingStartOn:
+                return false;
+            case SettingStartRememberLast:
+                return PlayerPrefs.GetInt(MicrophoneState, 1) == 1;
```
dooly123, line 125: I try to stay away from player prefs guess it makes sense here

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
BoatFloater: To be honest I thought so too but I was just following the existing pattern in the file. I've moved this and a bit more of the initialization into `BasisLocalMicrophoneIconDriver`.

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalMicrophoneDriver.cs
BoatFloater: The PlayerPrefs for this was already created and written to (but seemingly never read); I just consumed it for this feature.

PR #641, opened 2026-03-14
Add Cilbox whitelist to video player and render target as well as string encoding for urls

Mr-Zero88:


Basis/Packages/com.basis.shim/Shims/CilboxPropBasis.cs
```diff
@@ -104,6 +106,7 @@ public class CilboxPropBasis : Cilbox
 			"UnityEngine.UI.Text",
 			"UnityEngine.Vector3",
 			"UnityEngine.Vector4",
+			"UnityEngine.Video.VideoPlayer"
```
dooly123: we need todo a better job here, video players can be used to do a lot of nasty things we should build some way of parsing safe data into it.

dooly123 (changes requested):

Basis/Packages/com.basis.shim/Shims/CilboxPropBasis.cs
Mr-Zero88: Now using VideoPlayerShim

PR #659, opened 2026-03-19
Fix network id to align with the rest of basis for shim

Mr-Zero88:


BoatFloater:
What is your reasoning for reverting `LowLevelGetHierarchyPath`? The old logic is not sound.

Toys0125:
@BoatFloater Which part was not sound as it properly represents the entire hiearchy path, not just the gameobject, and multiple components on the same gameobject are already incremented with index. The code was sound we just were overwritting the code.
Only other thing we did was improve the client identifier to read out better when needing to diagnosis issues. 
Ex:`/BasisFramework/NetworkManagement/21676f849e9142a1b6eade54500570ae20260319/Screen Pickup:BasisObjectSyncNetworking0`

BoatFloater:
@Toys0125 My PR was 2 distinct features
* Improve the `LowLevelGetHierarchyPath` (not related to Cilbox)
* Use `LowLevelGetHierarchyPath` in BasisNetworkShim

The old `LowLevelGetHierarchyPath` does not account for poorly designed (but valid) hierarchy such as siblings with the same name; it also uses string concatenation which is not as efficient as StringBuilder. I also used `.FullName` for Type because I read somewhere that `GetType()` is not consistent between runtimes, though I read more and it's probably not an issue these days.

If adding BasisNetworkShim automatically sets its ID without needing an override, then just remove the override. Revert your change to `BasisNetworkBehaviour`.

Toys0125:
Your changes also gave weird formating errors of `[` inside the client identifier. I do believe that StringBuilder for concating arrays will be faster, but using string.concat which is the underlying + operator uses less memory.

Toys0125:
> The old `LowLevelGetHierarchyPath` does not account for poorly designed (but valid) hierarchy such as siblings with the same name;

Also we do not support having same named siblings at all. When we build we require that all gameobjects be named differently, which might be why it wasn't an issue.

BoatFloater:
@Toys0125 Those are not formatting errors; that is the sibling index. Please try to understand code before you change it.
I just built a scene with two networked siblings of the same name and it was allowed to build. Even though it's not recommended to do so in regular Unity, there's no reason not to account for it here as a safeguard.
Also, Basis is a framework; not every network component will be loaded through an addressable anyway.

Mr-Zero88:
No we understood the sibling index. just dont put it everywere. only if duplicate sibling are found index them but also not globaly but indexed localy only for same siblings.

BoatFloater:
@Mr-Zero88 Could you update your PR to make that change?

Basis/Packages/com.basis.shim/Shims/BasisShims.cs
```diff
@@ -40,8 +40,8 @@ public static BasisNetworkShim MakeNetworkable( MonoBehaviour mb )
 			bi = mb.gameObject.AddComponent<BasisNetworkShim>();
 			string setGUID = BasisNetworkBehaviour.LowLevelGetHierarchyPath(bi);
 
-			bi.AssignNetworkGUIDIdentifier(setGUID);
-			Debug.Log( $"ADDING ASSIGN: {bi} {setGUID}");
+			/* bi.AssignNetworkGUIDIdentifier(setGUID);
```
dooly123, line 43: why?

dooly123 (changes requested):

PR #660, opened 2026-03-19
Update network shims again

BoatFloater:
* Add `RequestOwnershipIfNone` which just shims `RequestWhoIsOwnershipAsync` as void instead of Task for Cilbox

Rework fixes from this PR: https://github.com/BasisVR/Basis/pull/659
* Remove override logic in `BasisShims`.`MakeNetworkable` because the base `BasisNetworkBehaviour` will already do it automatically when `AddComponent` is called
* Make `LowLevelGetHierarchyPath` not add any sibling index if it is already a unique sibling

Let me know if this works for you @Toys0125 @Mr-Zero88

### DIFF
--- a/Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
+++ b/Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
@@ -235,16 +235,44 @@ namespace Basis
             Component[] components = obj.gameObject.GetComponents(obj.GetType());
             int index = System.Array.IndexOf(components, obj);
 
-            pathBuilder.Append($"{obj.gameObject.name}[{obj.transform.GetSiblingIndex()}]_{obj.GetType().FullName}_{index}");
+            pathBuilder.Append($"{obj.gameObject.name}{SiblingIndexIfNeeded(obj.transform)}:{obj.GetType().FullName}_{index}");
             Transform current = obj.transform.parent;
 
             while (current != null)
             {
-                pathBuilder.Insert(0, $"{current.name}[{current.GetSiblingIndex()}]/");
+                pathBuilder.Insert(0, $"{current.name}{SiblingIndexIfNeeded(current)}/");
                 current = current.parent;
             }
 
             return pathBuilder.ToString();
+        }
+        private static string SiblingIndexIfNeeded(Transform t)
+        {
+            Transform parent = t.parent;
+            string name = t.name;
+            if (parent == null)
+            {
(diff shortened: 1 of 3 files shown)


Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
```diff
@@ -235,17 +235,45 @@ public static string LowLevelGetHierarchyPath(BasisNetworkContentBase obj)
...
+                {
+                    if (go != t.gameObject && go.name == name)
+                    {
+                        return $"[{t.GetSiblingIndex()}]";
+                    }
+                }
+            }
+            else
+            {
+                int childCount = parent.childCount;
```
Toys0125, line 265: wouldn't transform.find be faster in this block?

Basis/Packages/com.basis.shim/Shims/BasisNetworkShim.cs
```diff
@@ -47,5 +47,10 @@ public override void OnPlayerJoined(BasisNetworkPlayer player)
         {
 			PlayerJoined?.Invoke( player );
         }
+
+        public void RequestOwnershipIfNone()
+        {
+	        RequestWhoIsOwnershipAsync();
```
Toys0125, line 53: Needs testing as cilbox also tracks time a function takes and may timeout cilbox.

Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
BoatFloater: I avoided `Find` because it doesn't work if the name has "/" in it and it could find the target GO (instead of any number of siblings of the same name); better to iterate through all siblings to be sure

Basis/Packages/com.basis.shim/Shims/BasisNetworkShim.cs
BoatFloater: I tested it. Since I don't `await` here, it returns immediately

Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
dooly123: never use transform.find security risk

dooly123: ow actually, its not bad interesting i have learned something

PR #671, opened 2026-03-22
Fixed the master audio slider not affecting player voices

timothysoup:


dooly123:
no, we need to make sure we always have a catch all for the audio, the real reason the audio is not working with the master is we never do  in the player audio  playeraudio * listener volume

dooly123:
If you would like to fix it just need to multiply in the remote audio system

timothysoup:
#673

PR #673, opened 2026-03-22
Fixed the master audio not affecting player voices

timothysoup:


Basis/Packages/com.basis.framework/Networking/Recievers/BasisAudioReceiver.cs
```diff
@@ -579,10 +579,11 @@ private void ProcessNoResample(float[] data, int frames, int channels)
             Buffer.BlockCopy(segment, 0, _inputScratch, 0, frames * sizeof(float));
 
             float dampen = DirectionalDampeningMultiplier;
+            float listenerVolume = AudioListener.volume;
```
dooly123: why dont we just use the saved volume in the settings, accessing audiolistener.volume i suspect is slower then that

PR #687, opened 2026-03-25
Update OpusSharp to v1.6.3

SineVector241:


SineVector241:
Note: OpusSharp automatically switches to use the static version when inside an iOS runtime. Not sure if unity also provides the same runtime information.

PR #697, opened 2026-03-27
add loading scene

BoatFloater:
* Add BasisLoadingScene
  * Loads when another scene unloads if there are no scenes left
  * Has a floor and skybox & resets the main camera clear flags to skybox
* Move Initialization to `com.basis.framework`
  * It depends on prefabs from `com.basis.framework`


dooly123:
i was just thinking about this today

BoatFloater:
Ah nice; it's been in the back of my mind for a bit and I just had some time to take a crack at it

dooly123:
tested looks good, well done.

PR #699, opened 2026-03-27
Sphere preview of content

Toys0125:
## Summary
- add an opt-in storage setting for shared content previews
- render live avatar/prop/world previews inside content share spheres
- improve preview material fallback handling for generic renderers
- update the avatar orb assets to better present shared previews
- keep the default behavior unchanged by leaving previews disabled by default

## Testing
- Ran in Unity and did testing with setting on and off. Does not retro-active load previews for content shares already spawned.


Toys0125:
Unity_2026-03-27_01-53-38

dooly123:
I have issues with these changes:

I really think we just need to put in the time to also uploading as metadata that is additional data that we can grab when required. 

in this additional data we should include a imposter to display for this reason.

currently we could be creating a ticking time bomb of performance hurty or even crash things by doing this

dooly123:
feel free to open pull requests for parts that are not that of course :)

PR #700, opened 2026-03-28
Fix Pollution of Client Builds in CI

Toys0125:
When the headless saved the cache, it pollutes the cache with its build settings. So decided its best to reuse the client's cache but not save it. And added explicit build setting guard to the client build pipeline.


Toys0125:
Going to delete my branch to clear all the history after this.

PR #701, opened 2026-03-28
feat(sdk): add BasisParameterDriver for declarative animator parameter control

towneh:
## Summary

- Adds `BasisParameterDriver`, a `StateMachineBehaviour` that executes
  parameter operations (Set, Add, Random, Copy) on animator state entry.

- Adds `BasisParameterDriverEditor`, a custom inspector providing a
  card-based UI for managing operations with foldouts, reordering, and
  per-type field layouts.

## Motivation

Currently there is no SDK-level way for creators to declaratively modify
animator parameters in response to state transitions. Creators either
need to write custom MonoBehaviours or rely on complex animator setups
to achieve common patterns like:

- Setting a parameter to a fixed value on state entry
- Accumulating a value over repeated transitions (Add)
- Randomising int/float/bool parameters with configurable ranges and
  repeat prevention
- Copying one parameter to another with optional range remapping

`BasisParameterDriver` covers all of these as a drop-in
`StateMachineBehaviour`.

## What's included

| File | Purpose |
|------|---------|
| `BasisParameterDriver.cs` | Runtime behaviour — data model, execution logic, type-safe read/write helpers |
| `BasisParameterDriverEditor.cs` | Custom inspector — operation list UI with add/remove/reorder |

### Operation types

| Type | Description |
|------|-------------|
| **Set** | Write a constant value to the destination parameter |
| **Add** | Read current value, add a delta, write back |
| **Random** | Roll a random float/int/bool with min/max, chance, and repeat prevention |
| **Copy** | Read from a source parameter and write to destination, with optional range remapping |

### DIFF
--- a/Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
+++ b/Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
@@ -60,8 +60,12 @@ public class BasisParameterDriver : StateMachineBehaviour
     // Runtime state
     // ------------------------------------------------------------------
 
-    // Tracks previous int values per parameter for preventRepeats.
-    private readonly Dictionary<string, int> _lastIntValues = new Dictionary<string, int>();
+    // Tracks previous int values per parameter hash for preventRepeats.
+    private readonly Dictionary<int, int> _lastIntValues = new Dictionary<int, int>();
+
+    // Pre-computed parameter name hashes (populated once on first enter).
+    private int[] _destHashes;
+    private int[] _srcHashes;
 
     // ------------------------------------------------------------------
     // StateMachineBehaviour callbacks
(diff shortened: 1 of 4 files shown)


Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+    // Tracks previous int values per parameter for preventRepeats.
+    private readonly Dictionary<string, int> _lastIntValues = new Dictionary<string, int>();
+
+    // ------------------------------------------------------------------
+    // StateMachineBehaviour callbacks
+    // ------------------------------------------------------------------
+
+    public override void OnStateEnter(Animator animator, AnimatorStateInfo stateInfo, int layerIndex)
+    {
+        foreach (Operation op in operations)
```
dooly123: use a for loop instead of foreach here there is a performance difference when you store the ,length outside for the loop loop.

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+            Execute(animator, op);
+    }
+
+    // ------------------------------------------------------------------
+    // Execution
+    // ------------------------------------------------------------------
+
+    private void Execute(Animator animator, Operation op)
+    {
+        if (string.IsNullOrEmpty(op.destination))
```
dooly123: could we validate this code another way leading to always having valid data that then wouldnt require this check anymore?

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+            return;
+
+        AnimatorControllerParameterType destType = GetParamType(animator, op.destination);
+        if (destType == 0)
+        {
+            Debug.LogWarning($"[RNGParamDriver] Destination parameter '{op.destination}' not found on animator.", animator);
```
dooly123: BasisDebug.LogWarning Exists

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+                int result = UnityEngine.Random.Range(min, max + 1);
+
+                if (op.preventRepeats && range > 1 &&
+                    _lastIntValues.TryGetValue(op.destination, out int last) && last == result)
+                {
+                    result = min + ((result - min + 1) % range);
+                }
+
+                _lastIntValues[op.destination] = result;
+                animator.SetInteger(op.destination, result);
```
dooly123: use the hash converter else everytime you do this it has to find it

https://docs.unity3d.com/6000.3/Documentation/ScriptReference/Animator.StringToHash.html

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+
+    // ------------------------------------------------------------------
+    // Helpers
+    // ------------------------------------------------------------------
+
+    private static float ReadParamAsFloat(Animator animator, string name, AnimatorControllerParameterType type)
+    {
+        switch (type)
+        {
+            case AnimatorControllerParameterType.Float: return animator.GetFloat(name);
```
dooly123: yer the hash table will do better here

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
```diff
@@ -0,0 +1,210 @@
...
+            default: return 0f;
+        }
+    }
+
+    private static void WriteParam(Animator animator, string name, AnimatorControllerParameterType type, float value)
+    {
+        switch (type)
+        {
+            case AnimatorControllerParameterType.Float:
+                animator.SetFloat(name, value);
```
dooly123: and here

dooly123 (commented):
i could merge it currently as it does work but i think fixing the below issue will make it faster for everyone

dooly123:
once its merged i will add it to the avatar police content do you want this to work in worlds and scenes aswell?

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: Moving to hash table look ups should be much cleaner now to check for 0

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: Using

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: Now using a for loop

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: implemented

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: implemented

Basis/Packages/com.basis.sdk/Scripts/BasisParameterDriver.cs
towneh: noted and implemented

PR #703, opened 2026-03-31
Prevent Face Tracking from sending when disconnected from server.

Toys0125:
At the moment the script will continue to send to the network channel after disconnect from server, which will be null.

## Additions
* Adds a event action called OnLocalPlayerConnectionStateChanged
* LocalPlayerIsConnected invokes events

## Improvement
* Adds check if can send in face tracking.
* Adds event delegate function for state change.

### DIFF
--- a/Basis/Packages/com.basis.framework/Networking/BasisNetworkConnection.cs
+++ b/Basis/Packages/com.basis.framework/Networking/BasisNetworkConnection.cs
@@ -22,22 +22,7 @@ namespace Basis.Scripts.Networking
     {
         public static NetPeer LocalPlayerPeer { get; set; }
         public static NetworkClient NetworkClient { get; set; } = new NetworkClient();
-        private static bool _localPlayerIsConnected;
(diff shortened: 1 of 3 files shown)


Basis/Packages/com.basis.framework/Networking/BasisNetworkConnection.cs
```diff
@@ -22,7 +22,22 @@ public static class BasisNetworkConnection
...
+            get => _localPlayerIsConnected;
+            set
+            {
+                if (_localPlayerIsConnected == value)
+                {
+                    return;
+                }
+
+                _localPlayerIsConnected = value;
+                OnLocalPlayerConnectionStateChanged?.Invoke(value);
```
dooly123: im not a big fan of bool changes callback fires from that, 

its ever so slower then just calling = true & invoke(); seperately

dooly123: im also pretty sure we have a callback already

PR #704, opened 2026-03-31
Add multi-platform aware cache.

Toys0125:
Branch summary: Platform-aware cache for downloaded BEE/BME content.

This branch adds multi-platform awareness to the shared bundle cache so Linux and Windows builds no longer clobber each other when they download the same remote content. Previously the cache keyed on remote URL and UniqueVersion only, which meant avatars, props, worlds, preloads, and content-share items could reuse or overwrite the wrong platform payload. That would cause cases where metadata said a platform was supported but the cached bytes on disk were actually from another platform.

What changed:
- Added DownloadedPlatform to BEE metadata.
- Made BEE/BME cache filenames platform-aware.
- Made the on-disk metadata index platform-aware instead of URL-only.
- Prefer the current runtime platform when resolving cached content.
- Force a redownload when an older cached bundle is present but fails to load.
- Updated cache listing, eviction, deletion, and clear-cache flows to handle multiple platform variants of the same remote URL correctly.

Scope:
- Basis/Packages/com.basis.bundlemanagement/BasisBEEExtensionMeta.cs
- Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
- Basis/Packages/com.basis.bundlemanagement/BasisIOManagement.cs
- Basis/Packages/com.basis.bundlemanagement/BasisLoadhandler.cs
- Basis/Packages/com.basis.bundlemanagement/BasisStorageManagement.cs

Net result:
- A StandaloneLinux download no longer overwrites the StandaloneWindows cache entry for the same content, and vice versa.
- The fix applies across all content types that use the shared bundle cache, not just avatars.
- Older cache entries are still handled defensively through fallback path resolution and redownload-on-failure behavior.

### DIFF
--- a/Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
+++ b/Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
@@ -9,7 +9,7 @@ public static class BasisBeeManagement
     {
         return metaInfo != null &&
                !string.IsNullOrEmpty(metaInfo.DownloadedPlatform) &&
-               BasisBundleConnector.PlatformMatch(metaInfo.DownloadedPlatform);
+               BasisIOManagement.CachePlatformMatchesCurrent(metaInfo.DownloadedPlatform);
     }
 
     /// <summary>
--- a/Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
+++ b/Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
@@ -5,6 +5,13 @@ using System.Threading.Tasks;
 using UnityEngine;
 public static class BasisBeeManagement
 {
+    private static bool HasCompatibleDownloadedPlatform(BasisBEEExtensionMeta metaInfo)
+    {
+        return metaInfo != null &&
+               !string.IsNullOrEmpty(metaInfo.DownloadedPlatform) &&
+               BasisBundleConnector.PlatformMatch(metaInfo.DownloadedPlatform);
+    }
+
     /// <summary>
     /// this allows obtaining the entire bee file
     /// </summary>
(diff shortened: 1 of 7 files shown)


Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
```diff
@@ -63,10 +77,31 @@ public static async Task HandleBundleAndMetaLoading(BasisTrackedBundleWrapper wr
         try
         {
             AssetBundleCreateRequest bundleRequest = await BasisEncryptionToData.GenerateBundleFromFile(wrapper.LoadableBundle.UnlockPassword, output.Item2, output.Item1.AssetBundleCRC, report);
+            if (bundleRequest == null || bundleRequest.assetBundle == null)
+            {
+                if (shouldUseOnDiskMeta && !didForceRedownload)
```
dooly123, line 82: Will this cause a feedback loop of no bad bundle -> download -> attempt bad bundle -> download repeating

dooly123:
Would separating the folder structure also help to reduce churn of redownloading?

Basis/Packages/com.basis.bundlemanagement/BasisBeeManagement.cs
Toys0125: The next check of didForceRedownload is the check flag so it only does it once per invocation.

Toys0125: Only thing we could do is either delete the bad local file, and return an error about the download being bad.

Toys0125:
> Would separating the folder structure also help to reduce churn of redownloading?

Only redownload when it fails to load the cache. So atm it just leaves the old cache file.
So its `uniquefilename`-`platform`.BEE for the cache on new downloads only.

Toys0125:
> Would separating the folder structure also help to reduce churn of redownloading?

Granted using separate folders will reduce Startup Churn.

PR #706, opened 2026-03-31
Adds support for separate folder locations for each platform

Toys0125:
Doing this reduces the startup churn time, and we still support legacy cache files.

Additional changes to saved avatar data, AKA `LastUsedAvatar.BAS` to add the cache file that was used to reduce time it takes to load avatar from cache, and skips the cache warmup for loading your own avatar.

## Considerations
* We have full legacy support, and shouldn't have issues with files being placed into correct locations as we need full folder access to read and write files.
* If we read a legacy file we leave it as is.
* Legacy Lastused avatar is also fully supported until the user saves a new avatar.

## Improvements
* Remove wasteful lazyload as it just read through the files the same as out async task does but locks the game until it finds the one it was looking for.
* Using the direct file reference for the `LastUsedAvatar`

## Fixes
* Fixed the cache load stats breaking on legacy cache files.


dooly123:
after the next lts release we will merge

dooly123:
going to close this for now.... I need a compelling reason to merge thinking about it more

PR #709, opened 2026-04-02
Adding more network features for scripts. 

Toys0125:
* At the current moment we don't have direct way to assume if an object has no real owner easily.
For instance the server normally sends a remove owner and set the ID=0. Which works for everyone, except the first person. We start the player Id at 0. Which breaks this assumption with single player or low player count servers. So went and added a simple bool that the NetworkContent class will track if there is an owner assigned.
* Added NetworkTime, and NetworkDeltaTime to NetworkShim, to allow scripts to adjust delay based on it.


Toys0125:
Other option is when we release the owner we set `currentOwnedPlayer` to null. And make it `hasCurrentOwner => {currentOwnedPlayer != null}`

dooly123:
We really should just have a trygetownershipid that gives you back false or true,  done that a few places in the code base.

I like using zero as the starting point as well, that's programming

Toys0125:
> We really should just have a trygetownershipid that gives you back false or true, done that a few places in the code base.
> 
> I like using zero as the starting point as well, that's programming

Okay added the TryGetOwnerId in which checks if we have it cached, and outputs the owner id, otherwise returns false.

Basis/Packages/com.basis.shim/Shims/CilboxSceneBasis.cs
```diff
@@ -88,11 +88,7 @@ public class CilboxSceneBasis : Cilbox
 			"UnityEngine.DynamicGI",
 			"UnityEngine.Events.UnityAction",
 			"UnityEngine.Events.UnityEvent",
-			"UnityEngine.EventSystems.BaseEventData",
-			"UnityEngine.EventSystems.EventTrigger",
-			"UnityEngine.EventSystems.EventTrigger+Entry",
-			"UnityEngine.EventSystems.EventTrigger+TriggerEvent",
-			"UnityEngine.EventSystems.EventTriggerType",
+			"UnityEngine.EventSystems.*",
```
dooly123: im weary of .*  can see this becoming a magical ow shit moment if unity adds something

Basis/Packages/com.basis.framework/Networking/BasisNetworkBehaviour/BasisNetworkBehaviour.cs
```diff
@@ -302,6 +304,18 @@ public async Task<BasisOwnershipResult> RequestWhoIsOwnershipAsync(int Timout =
             BasisOwnershipResult Result = await BasisNetworkOwnership.RequestCurrentOwnershipAsync(clientIdentifier, Timout);
             return Result;
         }
+
+        public bool TryGetOwnerId(out ushort OwnerId)
```
dooly123, line 311: i need convincing this is better then just making people async the server and store the result

Toys0125: Its just a wrapper for the normal check you may do in normal C#, and this allows direct access for Cilbox scripts to ask for the OwnerId directly, reducing the amount of instructions needing to call.

PR #710, opened 2026-04-02
fix: possible race condition in StreamedAvatarFeature preventing data from being sent

kidkwazine:
StreamedAvatarFeature subscribed to PeerConnectedEvent in `Awake()` to set `_canSendNetworkData`, but Nethack guarantees this component is only created AFTER both avatar + network are ready so the event had likely already fired before the subscription existed (resulting in the flag being perma-false)

Replaced with a direct `if (BasisNetworkConnection.LocalPlayerPeer == null)` in `Update()`

Hoping this fixes the issue reported in disc server re: eye movements not being sent
(couldn't / didn't repro, but this stuck out to me when I looked into it to make sure it wasn't caused by my previous PR 😅 )


dooly123:
@Toys0125, I think we just need to instead of throwing code every frame to be evaluted lets just check the state on initalization, i dont want to touch this code since i dont have a way to test it.

dooly123:
look at the code nope this is a good solution

PR #724, opened 2026-04-09
feat: Update OpusSharp to v1.6.6 + binaries.

SineVector241:
Includes some fixes to the opus binaries and update OpusSharp to use the new v1.6.1 features, which should skip having to modify `StaticNativeOpus` from conditional compilation...


Toys0125:
So does this fix the problem of libopus.so being compiled as arm64? Or do I still need to do the link to opus.so?

SineVector241:
> So does this fix the problem of libopus.so being compiled as arm64? Or do I still need to do the link to opus.so?

Honestly not sure, seems pretty specific to unity itself as I've never had this issue before in my own projects which compiles to all platforms although I do use dotnet 9.0 so different runtime/compilation configuration IG...

Toys0125:
Easiest way is do a manual compile of Linux server from unity editor and check the dll after the compile. Under Linux_data/Plugins

PR #773, opened 2026-04-20
fix: update docs.basisvr.org links to correct paths

towneh:
## Summary
                                                                                                                                                                              
  - Added `/en` prefix to avatar, world, prop, and scene doc URLs                                                                                                             
  - Simplified the getting-started URL to the docs root                                                                                                                       
                                                                                                                                                                              
  ## Test plan                                                                                                                                                                
                                                                                                                                                                              
  - [x] Verify each link opens the correct documentation page in a browser


towneh:
Verified links open in the local repo.

One thing to note that when opening links from the Basis components you are challenged with:
{DB30FCBA-C4D3-42A8-94D1-BD50F50DF84E}

When opening links from the Basis Project Setup they open immediately.

PR #779, opened 2026-04-22
fix: ignore EditorOnly hierarchy in validator

hai-vr:
During avatar build, all GameObject hierarchy tagged with EditorOnly is removed from the avatar (before NDMF even runs). Therefore, only collect components pertaining to transforms that will exist during the build for the validation of Renderers, SkinnedMeshRenderers, and Transforms.

### DIFF
--- a/Basis/Packages/com.basis.sdk/Scripts/Editor/BasisAvatarValidator.cs
+++ b/Basis/Packages/com.basis.sdk/Scripts/Editor/BasisAvatarValidator.cs
@@ -303,18 +303,20 @@ public class BasisAvatarValidator
 
         ValidateTranslationDof(ref warnings, ref passes);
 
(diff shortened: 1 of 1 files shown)


Toys0125:
Putting my two cents, if you doing test-in-editor for the avatar those objects will still be there and could break the system if not properly accounted for...

Toys0125:
It would better if we did a build hook that properly accounts for those objects.

hai-vr:
> Putting my two cents, if you doing test-in-editor for the avatar those objects will still be there and could break the system if not properly accounted for...

Afaik Test in Editor also removes those EditorOnly objects

https://github.com/BasisVR/Basis/blob/developer/Basis/Packages/com.basis.sdk/Scripts/Editor/SDKInspector/BasisAvatarSDKInspector.cs#L663

Toys0125:
Okay don't see any other issue as the basis avatar for the "before" prefab is not running.

PR #786, opened 2026-04-28
Add third-person camera.

s-ilent:
This is a basic third person camera added to the BasisLocalCameraDriver. 

When third-person is enabled, the camera follows the player and collides with world geometry on specified layers. I also added some InputActions to handle switching between first/third person state, and zooming the camera in and out. Zooming will affect the camera distance and the FOV. The camera distance is also affected by the local player scale. 

The camera is updated on AfterSimulateOnLate to allow it to act as though it were detached from the player. 

The code probably needs some cleanup before it can be integrated.

Issues and Todos:
- Missing runtime settings menu to configure the camera parameters. 
- The call to ScaleheadToZero is skipped when in third-person mode. However, the head disappears when a mirror is on-screen. 
- SetupCollisionMask overrides the CameraCollisionMask and it probably shouldn't. 
- I haven't tested with avatars which are really small or really large before scaling.


Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+            }
+        }
+
+        /// <summary>
+        /// Orchestrates the third-person offset, scaling it by the avatar's height ratio
+        /// and performing a SphereCast to prevent the camera from clipping into geometry.
+        /// </summary>
+        private void SimulateThirdPerson()
+        {
+            if (!BasisDeviceManagement.IsUserInDesktop() || CameraInstance == null)
```
dooly123: CameraInstance will never be null?

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+
+            UpdateThirdPersonTargets(parentTransform, scale, dt);
+            ApplyThirdPersonTransform(scale);
+        }
+
+        private void SnapToFirstPerson()
+        {
+            if (!_wasThirdPerson) return;
+
+            transform.localPosition = Vector3.zero;
```
dooly123: transform.setlocalpositionandrotation instead

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+            transform.localRotation = Quaternion.identity;
+            CameraInstance.fieldOfView = DefaultCameraFov;
+
+            _wasThirdPerson = false;
+        }
+
+        private void UpdateThirdPersonTargets(Transform parentTransform, float scale, float dt)
+        {
+            Vector3 targetTrackingPos = parentTransform.position;
+            float targetPitch = parentTransform.rotation.eulerAngles.x;
```
dooly123: rotation. Euler converts it into a Euler (math), we then do that 2x

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+            if (!IsThirdPerson)
+            {
+                SnapToFirstPerson();
+                return;
+            }
+
+            Transform parentTransform = transform.parent;
+            if (parentTransform == null) return;
+
+            float dt = Time.deltaTime;
```
dooly123: we should be parsing down the delta time from BasisEventDriver

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+            float maxDistance = direction.magnitude;
+            float scaledRadius = CameraCollisionRadius * scale;
+
+            if (maxDistance > 0.001f && Physics.SphereCast(_currentCamParams.trackingPosition, scaledRadius, direction.normalized, out RaycastHit hit, maxDistance, CameraCollisionMask, QueryTriggerInteraction.Ignore))
+            {
+                desiredWorldPos = hit.point + (hit.normal * scaledRadius);
+            }
+
+            transform.position = desiredWorldPos;
+            transform.rotation = desiredRotation;
```
dooly123: SetPositionAndRotation

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
```diff
@@ -369,9 +399,192 @@ public void UpdateCameraScale(BasisHeightDriver.HeightModeChange HeightModeChang
...
+
+                _currentCamParams.pitch = DampAngle(_currentCamParams.pitch, targetPitch, RotationSmoothness, dt);
+                _currentCamParams.yaw = DampAngle(_currentCamParams.yaw, targetYaw, RotationSmoothness, dt);
+            }
+        }
+
+        private void ApplyThirdPersonTransform(float scale)
+        {
+            float zoomT = Mathf.InverseLerp(ThirdPersonMinZoom * scale, ThirdPersonMaxZoom * scale, _currentCamParams.distance);
+            CameraInstance.fieldOfView = Mathf.Lerp(ThirdPersonMinFoV, ThirdPersonMaxFoV, zoomT);
```
dooly123: this overrides users fov control are we sure we want todo this?

dooly123 (commented):
Please also update to the latest developer

dooly123:
thx for putting in the time for this we need todo a bunch more before its reliable, we also need to solve the distance for the reduction system being wrong hmm

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
s-ilent: Third-person FOV should be seperate from first-person FOV, I'll make it configurable with the other camera settings.

Basis/Packages/com.basis.framework/Drivers/Local/BasisLocalCameraDriver.cs
s-ilent: Oh okay, I wasn't sure

dooly123:
is this ready?

s-ilent:
Remaining issues:
- No settings in the menu yet
- deltaTime is still Time.deltaTime rather than from the event driver (I'm not really sure how to do that)
- Camera has some stuttering at low FPS (probably due to the update order)

If fixing those is a requirement then it isn't ready! Otherwise I think this is fine. I did some tests with different ways of setting up/smoothing the camera but I think this works the best.

PR #796, opened 2026-05-04
Desktop reticle

mriise:
## Summary
Reticle/Crosshair for desktop mode

adds:
- desktop reticle class (inner child of DesktopEye behavior)
- SDF reticle shader
- reticle object/quad placed at init
- setting toggle for the reticle (default off) with translations
- Material for the reticle in DesktopManagement

### Why
Selecting UI elements or attempting to grab a pickup that is either far away or too small without a reticle for the raycast origin on desktop is a chore in trial and error. This makes it infinitely easier to point/select a specific item, or at least be aware where the raycacst pointing is at.

## Required checks

- [x] **Transform access is combined and limited** — In hot paths, transform reads/writes go through `TransformAccessArray` or are otherwise batched. I have not added per-frame `transform.position` / `transform.rotation` / `transform.localPosition` calls inside loops.
- [x] **No new `GetComponent` / `AddComponent` where avoidable** — Where unavoidable, the result is cached on a field. None of these calls run inside `Update`, `LateUpdate`, `FixedUpdate`, jobs, or other per-frame code paths.

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [x] Tested with phone controls (mobile touch input)

- [x] Avatar swapping
- [x] Server swapping (joining / leaving / changing servers)

## Notes
Index, though this only ever applies to the desktop eye.


dooly123:
please explain what and why code also needs work

mriise:
whats not working?

dooly123:
please explain what and why, code also needs work but ill get to that once i understand the intent

mriise:
i updated the main description with what/why the same time as my last comment.

Toys0125:
tbh could also expand this for far away reticle for vr use as well. Many times when over 5 meters away have issues knowing exactly where the laser is hitting.

dooly123:
redical middle part should be transparent,    

  _quadGO.transform.localPosition = Vector3.forward * DistanceFromCamera;
            _quadGO.transform.localRotation = Quaternion.identity; 
just do the setlocalpositionandrotation


        Transform camTransform = BasisLocalCameraDriver.Instance.Camera.transform; 
bruh use the transform get 




        /// <summary>
        /// Screen-center aim reticle. The quad lives under the local camera so it
        /// tracks aim for free; lifecycle is driven by <see cref="Initialize"/> and
        /// <see cref="OnDestroy"/>. Toggle via <c>Reticle.SetEnabled(bool)</c>.
        /// </summary>
        [Header("Reticle")]
        public BasisDesktopReticle Reticle = new BasisDesktopReticle();

do we need to assign memory when its not in use?

mriise:
> tbh could also expand this for far away reticle for vr use as well. Many times when over 5 meters away have issues knowing exactly where the laser is hitting.

not sure if this applies to in vr. we already have the ring quad thing that gets placed on UI where the raycaster hits it if thats what you meant.

mriise:
> redical middle part should be transparent,

sure (it was slightly) but not fully imo, its an SDF so it might be interesting to add tweaks to it, but that can be someone else's adventure.
> _quadGO.transform.localPosition = Vector3.forward * DistanceFromCamera; _quadGO.transform.localRotation = Quaternion.identity; just do the setlocalpositionandrotation

yeh 
> ```
>     Transform camTransform = BasisLocalCameraDriver.Instance.Camera.transform; 
> ```
> 
> bruh use the transform get

also fair
> ```
>     /// <summary>
>     /// Screen-center aim reticle. The quad lives under the local camera so it
>     /// tracks aim for free; lifecycle is driven by <see cref="Initialize"/> and
>     /// <see cref="OnDestroy"/>. Toggle via <c>Reticle.SetEnabled(bool)</c>.
>     /// </summary>
>     [Header("Reticle")]
>     public BasisDesktopReticle Reticle = new BasisDesktopReticle();
> ```
> 
> do we need to assign memory when its not in use?

I mean its only ever on the DesktopEye so the only cost is the size of the object (which isnt much). We can defer initialization of the extra material and object until SetEnabled is triggered, but this is really shouldnt be a huge concern to begin with.

mriise:
Also its a bit silly to have Parent OF UI be not a good parent because its not at camera center.

dooly123:
if i had my way we wouldnt even be using transforms, there so slow

mriise:
need to parent to something unfortunately :\

Basis/Packages/com.basis.framework/Device Management/Devices/Desktop/BasisDesktopReticle.cs
```diff
@@ -0,0 +1,100 @@
...
+        private bool _userEnabled;
+        private bool _focused = true;
+
+        public void Initialize()
+        {
+            if (_quadGO != null) return; // idempotent
+
+            Transform camTransform = BasisLocalCameraDriver.Instance.transform;
+
+            _quadGO = GameObject.CreatePrimitive(PrimitiveType.Quad);
```
dooly123: i would prefer to use a gameobject over creating a primitive and then cleaning up that primitive

Basis/Packages/com.basis.framework/Device Management/Devices/Desktop/BasisDesktopReticle.cs
```diff
@@ -0,0 +1,100 @@
...
+            _quadGO.layer = LayerMask.NameToLayer("UI");
+            Object.Destroy(_quadGO.GetComponent<MeshCollider>());
+
+            _quadGO.transform.SetParent(camTransform, false);
+            _quadGO.transform.SetLocalPositionAndRotation(Vector3.forward * DistanceFromCamera, Quaternion.identity);
+            _quadGO.transform.localScale = Vector3.one * SizeMeters;
+
+            _material = new Material(SourceMaterial);
+
+            var renderer = _quadGO.GetComponent<MeshRenderer>();
```
dooly123: if its on disc everything below this would not need to be done

Basis/Packages/com.basis.framework/Device Management/Devices/Desktop/BasisDesktopEye.cs
```diff
@@ -71,6 +72,14 @@ public class BasisDesktopEye : BasisInput
         /// </summary>
         public float Z;
 
+        /// <summary>
+        /// Screen-center aim reticle. The quad lives under the local camera so it
+        /// tracks aim for free; lifecycle is driven by <see cref="Initialize"/> and
+        /// <see cref="OnDestroy"/>. Toggle via <c>Reticle.SetEnabled(bool)</c>.
+        /// </summary>
+        [Header("Reticle")]
+        public BasisDesktopReticle Reticle = new BasisDesktopReticle();
```
dooly123, line 81: we assign this to something but then below we have a null check?

Basis/Packages/com.basis.framework/Settings/SMModuleIcons.cs
```diff
@@ -24,6 +31,15 @@ public override void ValidSettingsChange(string matchedSettingName, string optio
             return;
         }
 
+        if (matchedSettingName == K_DESKTOP_RETICLE)
+        {
+            bool enabled = optionValue == "true";
+            // Reticle only exists while in desktop mode
+            BasisDesktopEye.Instance?.Reticle?.SetEnabled(enabled);
```
dooly123, line 38: since its the only thing for the .Instance would it make more sense to just make reticle a static getable to reduce complexity for the cpu

dooly123 (changes requested):

PR #797, opened 2026-05-04
feat: add AudioLink integration package

towneh:
## Summary

Adds two XR loader lifecycle events on `BasisDeviceManagement` and an opt-in
integration package, `com.basis.integration.audiolink`, that uses them.

- `BeforeLoaderTeardown` — fires before the active XR loader is deinitialised
  so subscribers can release state bound to the active XR-target graphics
  device.
- `AfterLoaderInitialized` — fires after a new XR loader has started its
  subsystems so subscribers can re-publish that state against the new
  graphics device.

Both events are raised by `BasisXRManagement` during the loader cycle via
internal raise methods on `BasisDeviceManagement`, keeping the public event
surface co-located with `OnBootModeChanged`.

`com.basis.integration.audiolink` subscribes to `AfterLoaderInitialized` and
`OnBootModeChanged` and toggles `AudioLink.AudioLinkEnabled` on enabled
instances after a mode switch to force a republish of `_AudioTexture`.
Without this, the global binding is wiped by the loader cycle and consumers
that resolve the texture by global lookup (custom URP passes, third-party
shaders) see a permanently null binding from the first frame after a
Desktop ↔ VR transition.

The integration package gates compilation on both `com.basis.framework` and
`com.llealloo.audiolink` (≥ 3.1.2) via asmdef `defineConstraints` +
`versionDefines`, so projects without either dependency are unaffected.

Refs: https://github.com/llealloo/audiolink/issues/365

## Required checks

- [x] **Transform access is combined and limited** — In hot paths, transform reads/writes go through `TransformAccessArray` or are otherwise batched. I have not added per-frame `transform.position` / `transform.rotation` / `transform.localPosition` calls inside loops.
- [x] **No new `GetComponent` / `AddComponent` where avoidable** — Where unavoidable, the result is cached on a field. None of these calls run inside `Update`, `LateUpdate`, `FixedUpdate`, jobs, or other per-frame code paths.

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes
Video showing editor testing workflow: https://youtu.be/EU3MV3vCIp0

- **Per-frame / `BasisEventDriver`:** `BasisAudioLinkRepublisher` is event-driven (`AfterLoaderInitialized`, `OnBootModeChanged`); no new `Update`/`LateUpdate`/`FixedUpdate` callbacks are added, so `BasisEventDriver` registration is not required.
- **`GetComponent` / scene scan:** the republisher uses `FindObjectsByType<AudioLink.AudioLink>(FindObjectsSortMode.None)` only on loader-init and boot-mode-change events — never per-frame. In scenes without an AudioLink instance the call returns an empty array and exits.
- **Jobification:** the work is property writes (`AudioLinkEnabled = false; AudioLinkEnabled = true;`) on Unity-managed objects, which must run on the main thread. No Burstable work is involved.
- **Avatar / server swap:** unaffected — the lifecycle events only fire on XR loader teardown / init.
- **Platforms / VR:** I'll fill in the boxes above once tested. The relevant scenario is a scene containing at least one `AudioLink` component, then triggering a Desktop ↔ VR hot-switch and confirming `_AudioTexture` stays bound on consumers.

### DIFF
--- a/Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
+++ b/Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
@@ -82,7 +82,6 @@ namespace Basis.Scripts.Device_Management.Devices
             }
 
             BasisDebug.Log($"Found Loader {result}", BasisDebug.LogTag.Device);
-            BasisDeviceManagement.RaiseAfterLoaderInitialized(result);
             StartDevice(result);
         }
 
@@ -104,7 +103,6 @@ namespace Basis.Scripts.Device_Management.Devices
             {
                 if (xRManagerSettings.isInitializationComplete)
                 {
-                    BasisDeviceManagement.RaiseBeforeLoaderTeardown(BasisDeviceManagement.StaticCurrentMode);
                     xRManagerSettings.DeinitializeLoader();
                 }
             }
--- a/Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
+++ b/Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
@@ -82,6 +82,7 @@ namespace Basis.Scripts.Device_Management.Devices
             }
 
             BasisDebug.Log($"Found Loader {result}", BasisDebug.LogTag.Device);
+            BasisDeviceManagement.RaiseAfterLoaderInitialized(result);
             StartDevice(result);
         }
 
@@ -103,6 +104,7 @@ namespace Basis.Scripts.Device_Management.Devices
             {
                 if (xRManagerSettings.isInitializationComplete)
                 {
+                    BasisDeviceManagement.RaiseBeforeLoaderTeardown(BasisDeviceManagement.StaticCurrentMode);
                     xRManagerSettings.DeinitializeLoader();
                 }
             }
--- a/Basis/Packages/com.basis.integration.audiolink/Runtime/BasisAudioLinkRepublisher.cs
+++ b/Basis/Packages/com.basis.integration.audiolink/Runtime/BasisAudioLinkRepublisher.cs
@@ -1,0 +1,42 @@
+using Basis.Scripts.Device_Management;
+using UnityEngine;
+
+namespace Basis.Integration.AudioLink
+{
+    /// <summary>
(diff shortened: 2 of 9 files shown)


towneh:
One last minute consideration here is that the separate package was created to follow a possible new naming convention for any third party integration requirements. If "integration" is too ambiguous and you would rather establish a preferred convention then I can easily change this. @dooly123

Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
```diff
@@ -82,6 +82,7 @@ public IEnumerator LoadXR()
             }
 
             BasisDebug.Log($"Found Loader {result}", BasisDebug.LogTag.Device);
+            BasisDeviceManagement.RaiseAfterLoaderInitialized(result);
             StartDevice(result);
```
dooly123, line 85: sooo why not just get on device change in start device and stop device? no need for another event

Basis/Packages/com.basis.framework/Device Management/Devices/Base/BasisXRManagement.cs
```diff
@@ -103,6 +104,7 @@ public void StopXR()
             {
                 if (xRManagerSettings.isInitializationComplete)
                 {
+                    BasisDeviceManagement.RaiseBeforeLoaderTeardown(BasisDeviceManagement.StaticCurrentMode);
```
dooly123: same here we dont need todo this StaticCurrentMode has callbacks already for when it changes

Basis/Packages/com.basis.integration.audiolink/Runtime/BasisAudioLinkRepublisher.cs
```diff
@@ -0,0 +1,42 @@
...
+            BasisDeviceManagement.OnBootModeChanged += OnBootModeChanged;
+        }
+
+        private static void OnLoaderInitialized(string loaderName) => Republish();
+
+        private static void OnBootModeChanged(string mode) => Republish();
+
+        private static void Republish()
+        {
+            global::AudioLink.AudioLink[] instances = Object.FindObjectsByType<global::AudioLink.AudioLink>(FindObjectsSortMode.None);
```
dooly123: will there ever been more then 1?

dooly123 (changes requested):

dooly123:
remind me tomorrow towneh and we can go over the naming

PR #806, opened 2026-05-06
Basisosc with vixxy support

Toys0125:
## Summary
Adds a scoped `BasisOsc` bridge for Cilbox scripts with OSC/OSCQuery runtime support and Vixxy variable pipeline integration.

- Added `Basis.Shims.BasisOsc` with exact, prefix, value, receive-all, unsubscribe, and publish APIs for Cilbox content.
- Added public OSC wrapper types (`OscData`, `OscDataKind`, `OscMessage`) and Cilbox whitelist/linker preservation for direct OSC use.
- Added OSC runtime service, OSCQuery node publishing, inbound OSC routing, and scoped publish/subscribe behavior for local avatars, remote avatars, props, and scenes.
- Mirrors numeric-compatible published values into Vixxy after publish address scoping passes, without exposing `HVRVariableStore` directly to Cilbox.
- Routes OSC receive draining through `BasisEventDriver` via `OSCAcquisitionServer.Simulate()` and avoids avoidable allocations in the main-thread dispatch path.
- Added a Cilbox OSC subscription sample and editor/debug visibility for OSC subscriptions.
- Added runtime tests covering subscriptions, scoped publishing, OSC type wrappers, OSCQuery value publishing, Vixxy mirroring, remote publish blocking, and SimpleOSC argument handling.

## Required checks

- 1* **No scene-wide discovery for dependencies** — New code is architected so it does not need `FindObjectOfType` / `FindObjectsOfType` / `GameObject.Find` / `FindGameObjectsWithTag` to locate what it depends on. References are wired in — registered through an existing manager/driver, injected at init, or passed in by the caller — rather than discovered by scanning the scene at runtime. If a scene scan is genuinely unavoidable, justify it under **Notes**.

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [x] Avatar swapping

## Notes
1. As Cilbox is runtime initialized scripts, and many of the networkshims are made at runtime, they don't have the proper references to pass onto, so they need to find their scope. I added caching so it only does it at init of the components.
- Public script callbacks that receive `OscMessage` still allocate wrapper objects by API design; the Vixxy/address dispatch path avoids per-frame owner-match allocations after initialization/warmup.
- Cilboxable sample scripts intentionally use normal `Debug.*` logging because Cilbox handles those logs correctly.
- Jobification was considered, but since most of the data is already on separate thread for SimpleOSC, jobification would not assist.

### DIFF
--- a/Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
+++ b/Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
@@ -125,7 +125,8 @@ namespace Basis.Shims.Samples
                 return report;
             }
 
-            for (int index = 0; index < arguments.Length; index++)
+            int argumentCount = arguments.Length;
+            for (int index = 0; index < argumentCount; index++)
             {
                 report += " | arg[" + index + "]=" + DescribeOscData(arguments[index]);
             }
@@ -185,7 +186,8 @@ namespace Basis.Shims.Samples
             }
 
             string description = string.Empty;
-            for (int index = 0; index < values.Length; index++)
+            int valueCount = values.Length;
+            for (int index = 0; index < valueCount; index++)
             {
                 if (index > 0)
                 {
--- a/Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
+++ b/Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
@@ -1,0 +1,226 @@
+using Cilbox;
+using Basis.Network.Core;
+using Basis.Scripts.BasisSdk;
+using HVR.Basis.Comms.OSC;
+using System.Text;
+using UnityEngine;
+
+namespace Basis.Shims.Samples
+{
+    [Cilboxable]
+    public class BasisOscCilboxSubscriptionExample : MonoBehaviour
+    {
+        private const string ExplicitTestAddress = "/avatar/parameters/test";
+        private const string ImplicitTestAddress = "test";
+
+        private Basis.Shims.BasisOsc osc;
+        private BasisAvatar avatar;
+        private Basis.Shims.BasisNetworkShim networkShim;
+        private bool isNetworkReady;
+
+        private void Start()
+        {
+            if (osc == null)
+            {
+                osc = GetComponent<Basis.Shims.BasisOsc>();
+                if (osc == null)
+                {
+                    Debug.LogError("BasisOscCilboxSubscriptionExample requires a BasisOsc component.");
+                    return;
+                }
+            }
+
+            avatar = BasisAvatar.GetGameObject(this).GetComponent<BasisAvatar>();
+            if (avatar == null)
+            {
+                Debug.Log("BasisOscCilboxSubscriptionExample did not find a BasisAvatar in the parent hierarchy.");
+            }
+
+            networkShim = this.gameObject.GetComponent<Basis.Shims.BasisNetworkShim>();
+
+            networkShim.NetworkReady += OnNetworkReady;
+            networkShim.NetworkMessageReceived += OnNetworkMessageReceived;
+
+            osc.Subscribe(ExplicitTestAddress, OnExplicitTestTriggered, out string ExplicitAddress);
+            osc.Subscribe(ImplicitTestAddress, OnImplicitTestTriggered, out string ImplicitAddress);
+
+            Debug.Log("Subscribed to " + ExplicitAddress + " and implicit \"" + ImplicitAddress + "\" on avatar local=" + avatar?.IsOwnedLocally + ".");
+        }
+
+        private void OnDestroy()
+        {
+            if (osc != null)
+            {
+                osc.Unsubscribe(ExplicitTestAddress, OnExplicitTestTriggered);
+                osc.Unsubscribe(ImplicitTestAddress, OnImplicitTestTriggered);
+            }
+
+            if (networkShim != null)
+            {
+                networkShim.NetworkReady -= OnNetworkReady;
+                networkShim.NetworkMessageReceived -= OnNetworkMessageReceived;
+            }
+            Debug.Log("BasisOscCilboxSubscriptionExample destroyed and unsubscribed from OSC and network events.");
+        }
+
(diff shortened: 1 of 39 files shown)


Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+
+            avatar = BasisAvatar.GetGameObject(this).GetComponent<BasisAvatar>();
+            if (avatar == null)
+            {
+                Debug.Log("BasisOscCilboxSubscriptionExample did not find a BasisAvatar in the parent hierarchy.");
+            }
+            // Cilbox will add the Shim when using GetComponent.
+            networkShim = this.gameObject.GetComponent<Basis.Shims.BasisNetworkShim>();
```
dooly123, line 39: use try getcomponent

Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
```diff
@@ -0,0 +1,228 @@
...
+                osc.Unsubscribe(ExplicitTestAddress, OnExplicitTestTriggered);
+                osc.Unsubscribe(ImplicitTestAddress, OnImplicitTestTriggered);
+            }
+
+            if (networkShim != null)
+            {
+                networkShim.NetworkReady -= OnNetworkReady;
+                networkShim.NetworkMessageReceived -= OnNetworkMessageReceived;
+            }
+            Debug.Log("BasisOscCilboxSubscriptionExample destroyed and unsubscribed from OSC and network events.");
```
dooly123, line 63: Debug is forbidden from usage use basisdebug

Basis/Packages/com.basis.shim/Shims/BasisOsc.cs
```diff
@@ -0,0 +1,1066 @@
...
+            public int ExactValueCallbackCount { get; internal set; }
+            public int PrefixCallbackCount { get; internal set; }
+            public int PrefixValueCallbackCount { get; internal set; }
+            public string[] ExactSubscriptions { get; internal set; } = Array.Empty<string>();
+            public string[] PrefixSubscriptions { get; internal set; } = Array.Empty<string>();
+            public string[] ExactRegistrationLines { get; internal set; } = Array.Empty<string>();
+            public string[] PrefixRegistrationLines { get; internal set; } = Array.Empty<string>();
+        }
+
+        private readonly Dictionary<string, OscMessageEvent> exactCallbacks = new Dictionary<string, OscMessageEvent>(StringComparer.Ordinal);
```
dooly123: is there a better storage method instead of dictonarys?

Basis/Packages/dev.hai-vr.basis.comms/Scripts/Systems/Runtime/OSC/Lyuma/SimpleOSC.cs
```diff
@@ -37,20 +37,51 @@ public class SimpleOSC
 {
 	public enum Impulse {IMPULSE}
```
dooly123, line 38: any reason lyuma name is used here for simple osc, should we make it a package instead?

Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
Toys0125: Dooly this is a cilbox script, Getcomponent is the standard way the TypeOverrider/AddComponent works. If you want to add support for TryGetComponent for cilbox do so.

Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
Toys0125: Cilbox scripts automatically have scope replaced by the proper BasisDebug log.

Basis/Packages/dev.hai-vr.basis.comms/Scripts/Systems/Runtime/OSC/Lyuma/SimpleOSC.cs
Toys0125: @dooly123 that's an @hai-vr question since they put the name of Lyuma probably from the Maintainer of SimpleOSC.

Basis/Packages/com.basis.shim/Shims/BasisOsc.cs
Toys0125: Add a different method using a class with two dictionaries instead. https://github.com/BasisVR/Basis/pull/806/commits/de6354d1dc373f81c251491c60a820c8548d997c

Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
dooly123: thx for checking for me

Basis/Packages/com.basis.shim/Samples/OSC Subscription Example/BasisOscCilboxSubscriptionExample.cs
dooly123: perfect

Basis/Packages/dev.hai-vr.basis.comms/Scripts/Systems/Runtime/OSC/Lyuma/SimpleOSC.cs
dooly123: im happy to merge it now with the goal if packaging it up later

hai-vr: SimpleOSC comes from a single-file script that Lyuma made ages ago, and is notably used here:
- https://gist.github.com/lyuma/120d2736d6963460fc641fe24c1b02f7
- https://github.com/lyuma/Av3Emulator/blob/master/Runtime/Scripts/A3ESimpleOSC.cs

PR #827, opened 2026-05-18
Steam transport Integration

saint4626:
## Summary
This PR adds a new embedded `com.basis.steamtransport` package for Steam networking support on top of the new multi-network stack.

The package registers a `steam` network stack and adds Steam lobby creation/joining, Steam relay transport, invite handling, lobby metadata, and a Steam Lobbies panel in the Basis menu. The existing LiteNetLib/default networking path is left intact.

I kept changes outside the Steam package as small as possible. The only shared Basis changes are:
- an `OnUpdate` hook in `BasisEventDriver` so Steam callbacks and transport polling can run through the existing event driver
- small public bridge methods/constructors in `BasisNetworkShell` so a custom `NetManager` can raise network events without reflection
- the Unity package lock entry for the embedded Steam package

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [x] Avatar swapping
- [x] Server swapping (joining / leaving / changing servers)

## Notes
Tested on Windows in the Unity Editor and in a local build.

I tested both desktop/non-VR mode and VR mode.
VR headset: Pico 4.

I also checked the main runtime flows I expected this to affect:
- Steam transport registers as the `steam` network stack
- avatar swapping works
- porps works spawns
- desktop/VR hot-switching still works

Steam App ID note: `BasisSteamSettings` currently defaults to Steam AppId `480` for development/testing. For the real Basis Steam build, this should be changed to the actual Basis Steam App ID so lobbies, invites, and relay all run under the shipped app.

Jobification note: I do not think this work is a good fit for Unity Jobs. The new work is mostly Steamworks callbacks, lobby state, transport event bridging, and Basis network lifecycle calls, which need to stay on the managed/main-thread side.

Hot-path note: Steam callbacks and transport polling are routed through `BasisEventDriver.OnUpdate`. Receive processing is bounded, and transport trace logging is opt-in through `BasisSteamSettings.EnableTransportTrace`.


uyjulian:
May I suggest https://github.com/ValveSoftware/GameNetworkingSockets so the networking under this backend can be tested without needing Steamworks API

saint4626:
> May I suggest https://github.com/ValveSoftware/GameNetworkingSockets so the networking under this backend can be tested without needing Steamworks API

I agree this could be useful for transport-level testing without requiring Steamworks initialization/AppId.

For PR I kept the scope focused on the actual Steamworks path because the goal is Steam lobbies, invites, and Steam relay integration. GameNetworkingSockets sounds like a good follow-up as a separate test/non-Steam backend, but I would not want to mix that into this PR and expand the scope.

dooly123:
finally getting to this: 

public static Action OnUpdate; points at a larger issue with the steam intergration package

that its tied down to the main thread and consumes time on main thread

any reason its on main thread?

dooly123:
I have merged the function additions
image

saint4626:
> finally getting to this:
> 
> public static Action OnUpdate; points at a larger issue with the steam intergration package
> 
> that its tied down to the main thread and consumes time on main thread
> 
> any reason its on main thread?

I kept it on the main thread because Steamworks callbacks run on the caller thread, and the current callback path touches lifecycle state

i think the architecture could be improved in future

dooly123:
oki in that case is there a better spot for the event?

saint4626:
> oki in that case is there a better spot for the event?

cleaner path might be to call BasisNetworkStackRegistry.TickActive from the existing network update path and remove the extra OnUpdate hook.
The only part I’m less sure about is SteamClient.RunCallbacks, because that also drives lobby/invite callbacks before there is necessarily an active network connection.

PR #839, opened 2026-06-01
feat: VR camera fly mode (RC Mode 2 twin-stick controls)

towneh:
## Summary

Adds a VR "fly mode" to the handheld capture camera in to  provide a  much greater degree of freedom for film-making, streaming and taking photos, using **RC "Mode 2"** twin-stick controls.

Fly Mode activate and recall is on the **B/Y button**, replacing the old thumbstick-*click* toggle. The thumbstick toggle was one motivation: repeated stick-clicks wears the thumbstick over time and presents a negative user association.

Proposal: Hold the camera, quick press B/Y → it detaches to world space and both hands pilot. quick press B/Y again → it returns to hand.

**Twin-stick mapping (RC Mode 2):**

| Axis | Stick | Effect |
| --- | --- | --- |
| Throttle | pilot Y | Altitude — straight world-up/down, heading-independent |
| Yaw | pilot X | Rate-based turn; heading **holds** when the stick re-centres |
| Pitch | free Y | Forward / back, relative to heading |
| Roll | free X | Strafe, relative to heading |

While flying, the avatar is pinned in place via look / movement / crouch locks.

Built on the existing fly infrastructure (`CameraPinSpace.WorldSpace`, the pin constraint, and the existing fly-speed / acceleration / momentum fields) rather than new scaffolding — only a `yawRate` tunable is added. 

Desktop fly mode is unchanged.

### Two commits

- **`fix: gate VR turn on the LookRotation lock`** — the snap/smooth-turn block applied controller turn input unconditionally, so a held `LookRotation` lock suppressed *desktop* look but not *VR* stick-turn. Added a lock guard mirroring the existing `MovementLock` guard. Without this, the pilot stick's yaw would also rotate the avatar while flying. This is a small change to a shared control path (`BasisLocalCharacterDriver`), so it is called out separately.
- **`feat: VR camera fly mode with RC Mode 2 twin-stick controls`** — the fly mode itself, entirely within `BasisHandHeldCameraInteractable`.

## Required checks

- [ ] **Tested** — I built and ran this locally. The change works in the editor and (where relevant) in a built player.

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

**Draft — not merge-ready yet.** open items:

1. **Not yet built locally**

2. **Known interaction: left-hand B/Y also opens the hamburger menu.** The hamburger menu is bound to the left-hand secondary button (`BasisActionDriver`), so launching/recalling with the *left* hand as pilot also toggles the menu. Right-hand piloting is clean. This is intentionally **not** fixed here — the fix is tracked in https://github.com/BasisVR/Basis/pull/838

**Checklist context:**
- **Jobification** — N/A. Single-camera flight integration on the local player only; a handful of vector ops per frame, not job-worthy.
- **Camera access** — operates on the handheld camera's own `captureCamera` (the prop's capture camera), not the local player camera, so `BasisLocalCameraDriver` discovery does not apply.
- **BasisEventDriver** — per-frame work runs via the existing `BasisLocalPlayer.AfterSimulateOnLate` registration; no new `Update`/standalone tick added.
- **Hot path** — `MoveVRFlying` runs per frame; it allocates nothing (`Vector3`/`Quaternion` are value types) and logs nothing. The only `BasisDebug` calls are on the launch/recall edges.

### DIFF
--- a/Basis/Packages/com.basis.framework/Camera/BasisHandHeldCameraInteractable.cs
+++ b/Basis/Packages/com.basis.framework/Camera/BasisHandHeldCameraInteractable.cs
@@ -49,6 +49,10 @@ public abstract class BasisHandHeldCameraInteractable : BasisPickupInteractable
     [Range(5f, 25f)]
     public float rotationSmoothing = 15f;
 
+    /// <summary>VR fly-mode yaw turn rate at full pilot-stick deflection (degrees/second). Rate-based:
+    /// the heading holds when the stick re-centres.</summary>
+    public float yawRate = 90f;
+
     [Header("Cinematic Controls")]
     /// <summary>Whether to use momentum/inertia for movement.</summary>
     public bool useMomentum = true;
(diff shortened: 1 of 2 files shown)


towneh:
This became stale and likely won't be revisited by me for the foreseeable future.

PR #856, opened 2026-06-12
feat(mediaplayer): multichannel surround audio via per-channel AudioSources

towneh:
## Summary

Adds **multichannel (e.g. 5.1) surround audio** to `com.basis.mediaplayer`: a stream's audio channels can each be played through their own positioned Unity `AudioSource`, so surround content is spatialised channel-by-channel in the world rather than collapsed to stereo.

**Demo (5.1 streamed live, channels positioned in-world):**

https://github.com/user-attachments/assets/ab2b0a63-4aba-4b6a-8c7a-262efa8b38ce

**What it adds**

- A `UnityMultiChannelSources` value on the existing `BasisAudioRouting` enum (the routing seam that was already scaffolded for this).
- `BasisMediaPlayerMultiChannelAudio` — a sink that holds a flat `Outputs` list of `AudioSource`s. Each output carries a `BasisMediaAudioChannel` component selecting its channel (Channel 1–8, or a stereo downmix). Place the AudioSources wherever the speakers should sit; the package owns playback, the consumer owns positioning.
- `BasisMultiChannelPcmSplitter` — broadcasts the native engine's interleaved PCM ring through a rolling per-channel window so every output reads at its own cursor via a small mix matrix. This means **the same channel can feed two AudioSources in different places** (graceful surround in two areas), and a **Stereo** output can mix a layout-aware downmix of the whole stream. An output whose selected channel isn't present in the stream (e.g. a 5.1 output fed a stereo source) stays silent.
- `BasisMediaPlayer` routes the PCM ring, format and volume to whichever sink the routing selects; the stereo `BasisMediaPlayerAudio` path is untouched. `LoadUrl`/`LoadLocalPath` now preserve the player's configured routing instead of letting the default convenience source reset it.

**Native (Windows)**

The Media Foundation AAC decoder now selects the output type whose channel count matches the input, keeping the discrete surround channels rather than landing on the decoder's stereo fold-down. The prebuilt `basis_media_native.dll` is rebuilt (Release) to include it.

**Editor / example**

Both new components get inspectors matching the shared Basis Media Player card styling, and `MediaPlayerMultiChannelStreaming.prefab` is a ready-to-use 5.1 example. `BasisMediaPlayerDiagnostics` reports the active sink's metrics (including engine-reported vs sink channel count), so the debug CSV works for either routing.

**Update — review response and A/V sync work**

Five follow-up commits since the original push:

- `d4f5bb36e` addresses review feedback: runtime `AudioRouting` changes re-route the PCM ring, a missing multichannel component warns instead of silently falling back to stereo, splitter taps are fully bounds-checked, and the diagnostics columns that don't apply to this sink are documented.
- `f742961c2` generalises channel handling: selections are now generic **Channel 1–8** (channels are content lanes, not just speaker positions), the stereo downmix folds any layout (3.0/4.0/5.0/5.1/7.1) with clip-safe normalisation, the sink waits for the decoder-reported format before building clips, outputs start sample-aligned via a shared `PlayScheduled`, and the demux maps AAC `channelConfiguration 7` to its real 8-channel count. The in-box Windows AAC decoder rejects 8-channel input (verified directly against the MFT), so 7.1 degrades to video-with-silent-audio there; platforms whose decoders accept it get the correct count.
- `eff907311` + `59990ed2a` add **PTS-scheduled audio release** on Windows: decoded PCM now carries presentation timestamps through the ring and is released against the same clock video presents on, so connect-burst backlogs can't accumulate as a permanent audio delay, ring drops are whole-frame (channel order survives overflow), and the jitter buffer floors at the audio pipeline depth when audio is present. The gate clock is low-passed so segmented transports (HLS) anchor on the mean of their segment-cadence wobble. Lip-sync verified side-by-side against VLC on a 5.1 TS stream and a Twitch HLS stream; glass-to-glass latency on audio-bearing streams is ~0.46 s as a result. The committed DLL is rebuilt from the branch tip.
- `b5178ca85` adds **LPCM audio over m2ts transport streams**: Blu-ray HDMV LPCM (stream_type 0x80, 192-byte packets) carries uncompressed PCM with no platform decoder involved, so **any channel count up to 7.1 plays on every platform — including Windows**, whose in-box AAC decoder caps at 5.1. The demuxer auto-detects the m2ts packet stride; the Windows decode layer converts/reorders the PCM straight into the PTS-scheduled ring. Producible with stock ffmpeg (`-c:a pcm_bluray -f mpegts -mpegts_m2ts_mode 1`); ~6 Mbps for 7.1/48k — venue/LAN delivery. The example prefab gains Channel 7/8 side speakers, making it a full 7.1 example. Verified: 7.1 LPCM with all eight channels audible on Windows, 5.1 LPCM (24-bit path), AAC regression clean.

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

**Verification.** Tested end-to-end on Windows against live streams: 5.1 (channels confirmed by ear to land in the right positions) and stereo, across **MPEG-TS over HTTP, RTSP/RTSPT, and HLS**, on both the multichannel prefab and the original stereo prefab. The stereo `BasisMediaPlayerAudio` path is unchanged (verified no regression). The `BasisMultiChannelPcmSplitter` de-interleave / carry / fan-out / downmix logic also has a standalone correctness test (frame-misaligned source, multiple readers, ITU downmix).

**Coordination with #853 (RIST).** This branch shares `BasisMediaPlayer.cs` with #853 but in disjoint regions — it trial-merges clean, no conflict either order. The committed `basis_media_native.dll` here is a non-RIST build; if both land, a rebuild from merged source keeps the binary current (no functional overlap — RIST is opt-in and off in the default build).

**Why most checklist boxes are N/A.** This is an audio-routing feature — it touches no transforms, camera, Addressables, or scene discovery, and adds nothing to `BasisEventDriver`, so those boxes are N/A and ticked per the template instruction.

- **`BasisEventDriver` (per-frame work).** The sink uses a MonoBehaviour `Update`, matching the existing `BasisMediaPlayerAudio` sink and `BasisMediaPlayer` itself (the package doesn't currently route audio through `BasisEventDriver`). The `Update` body is a cheap format-change poll (reads two volatile ints, rebuilds only when the stream's rate/channels change) — no per-frame allocation or logging. Happy to migrate the audio sinks to `BasisEventDriver` if you'd prefer; kept it consistent with the sibling for now.
- **Jobification.** The de-interleave runs on Unity's audio thread via the streaming `AudioClip` PCM callbacks, which isn't a Job/Burst context — N/A.
- **`GetComponent`.** The one component lookup (an output's `BasisMediaAudioChannel`) uses `TryGetComponent` and runs only on a (re)build when the stream format changes, never per-frame.
- **Hot paths.** The audio-thread paths (`ReadMixed`, the splitter pump, the per-output callback) allocate nothing, log nothing, and cache `.Length` into locals; the single warning log is on the build path.

**Not in this PR (deliberately):**

- **Android (Quest).** The native backend (AMediaCodec) already surfaces the real channel count, so the wiring carries over, but I haven't done a device build/run yet — leaving Android unticked rather than guessing.
- **44.1 kHz sources.** While testing I found a pre-existing, player-wide issue: a streaming `AudioClip` is consumed at the project output rate (48 kHz), so any non-48 kHz source plays slightly sharp and over-drains the buffer. It affects the stereo `BasisMediaPlayerAudio` identically and isn't specific to this change, so I'll file it separately rather than fold a resampler in here.
- **Flows.** I didn't specifically exercise a runtime desktop↔VR swap during multichannel playback; the feature adds no XR/avatar/networking logic, but flagging it for honesty.


towneh:
﻿Since the original push this branch has grown five follow-up commits — a quick map for anyone reviewing:

**Review feedback** — a contributor shared some detailed review feedback with me outside the thread; summarising here so the responses are on record. Taken (in `d4f5bb36e` / `f742961c2`): runtime `AudioRouting` changes now re-route the PCM ring; requesting multichannel without the component warns instead of silently falling back to stereo; splitter taps are fully bounds-checked; the sink waits for the decoder-reported format before building clips; all outputs start sample-aligned via one shared `PlayScheduled`; the stereo downmix is normalised so a full-scale input can't clip; each binding pins its splitter; the MF output-type ranking prefers a channel-count match over float so the decoder's stereo fold-down can't outrank discrete 5.1. Not taken, with verification: the `Clock.ExternalSource` concern (the clock only follows a source reporting `HasMediaTime`, which neither sink does on the native path), prefab `PlayOnAwake` (no-op — no clip exists until the sink builds one), a comment-encoding query (the files are clean UTF-8), an audio-callback lock redesign (reading the ring from the PCM callback is the package's existing contract, and the lock only copies floats), and the inspector namespace / `BindByName` dedup suggestions (the two new inspectors match the package's six existing ones — better as a package-wide style pass).

**Generic channels** (`f742961c2`) — selections are now **Channel 1–8** rather than 5.1-specific labels (channels are content lanes, not just speaker positions), the stereo downmix folds any layout (3.0/4.0/5.0/5.1/7.1), and the demux maps AAC `channelConfiguration 7` to its real 8-channel count. Worth knowing: the in-box Windows AAC decoder rejects 8-channel input outright (probed directly against the MFT), so 7.1-over-AAC plays video with silent audio on Windows — which led to the last commit below.

**PTS-scheduled audio** (`eff907311` + `59990ed2a`) — decoded audio previously free-ran, so any backlog (e.g. the burst a TS server flushes at connect) persisted as a permanent audio-behind-video offset, and ring overflow could rotate the channel order. Audio now carries presentation timestamps through the PCM ring and is released against the same clock video presents on, with all drops whole-frame multiples. The jitter buffer floors at the audio pipeline depth when audio is present, so glass-to-glass on audio-bearing streams is ~0.46 s — about AVPro parity, and well under VLC for the same sync. The gate clock is low-passed so segmented transports (HLS) anchor on the mean of their segment-cadence wobble rather than a random phase.

**LPCM over m2ts** (`b5178ca85`) — Blu-ray HDMV LPCM (stream_type 0x80, 192-byte packets, auto-detected) carries uncompressed PCM with no platform decoder involved, so **any channel count up to 7.1 plays on every platform, including Windows**. Producible with stock ffmpeg (`-c:a pcm_bluray -f mpegts -mpegts_m2ts_mode 1`); ~6 Mbps for 7.1/48k, aimed at venue/LAN delivery. The example prefab now has all eight channel speakers.

Verified across the work: 5.1 and 7.1 channel placement plus lip-sync side-by-side with VLC, stereo regression on TS/RTSPT/HLS, reconnect cycles, and the splitter test suite. The committed DLL is rebuilt from the branch tip.

towneh:
@dooly123 this is the final version of what I was demoing earlier.

towneh:
Docs updated https://docs.basisvr.org/en/docs/world/media-player

towneh:
I received further code review feedback directly which has been addressed as follows (in `7afbf06e9`):

- **Unsupported LPCM sample rates / bit depths were announced by the demuxer, then silently dropped by the decoder.** The TS demuxer now only announces 48 kHz / 16- or 24-bit LPCM, so a 96/192 kHz or 20-bit stream degrades to silence with video intact (the same graceful behaviour as unsupported AAC) instead of half-configuring the audio sink.

- **`PcmRing::write` truncated an over-capacity write without advancing the timestamp.** It now drops whole frames and carries the PTS forward, so the retained samples keep a correct timestamp and the channel order isn't rotated by a sub-frame trim.

- **The 16-bit AAC drain path wrote arbitrary 4096-float chunks, splitting interleaved frames.** It now writes whole interleaved frames only, keeping the ring's per-chunk PTS exact.

- **The LPCM channel remap had no spec reference.** Added a comment citing ffmpeg's `pcm_bluray` remap for the 5.1/7.0/7.1 assignments; the mapping was also verified by ear against a 7.1 channel-marker stream.

- **Smaller items:** `editor-extension-mode="True"` on both inspector UXML files, `formatKnown` made volatile to match the format pendings beside it, and a redundant clip null-check removed.

- **Audio-thread contract:** added a comment noting `IBasisPcmSource.ReadPcm` must be non-blocking and allocation-free, since it runs under the lock from Unity's PCM callback.

One flagged item I didn't change: the `Pump()` negative-index concern doesn't occur in practice — `carryLen` is always a sub-frame remainder (`< channelCount`) and `frames >= 1`, so `usable >= channelCount > carryLen` and the read index stays `>= 1`. I added a comment documenting that invariant rather than altering correct code. The `dspTime + 0.05` scheduling lead and the `float[][]` window layout were left as-is per the accompanying notes.

PR #871, opened 2026-06-14
feat(mediaplayer): split-stream, on-demand delivery, and optional page-URL resolution

towneh:
## Summary

Several related additions to the media player, plus an optional integration package, that together let it play high-resolution adaptive sources (YouTube/Twitch and similar) — not just the live broadcast URLs it handles today.

**1. Split-stream playback.** Adaptive sources above ~360p serve video and audio as *separate* streams. `basis_media_open_dual` runs a second demux thread for an audio-only URL into the same decoder, so both present in sync on one clock. Exposed via `BasisMediaSource.AudioUri` (null = single muxed stream, unchanged).

**2. Live vs on-demand, auto-detected.** `BasisMediaSource.Delivery` (`Auto`/`Live`/`OnDemand`) selects the playback clock. `Auto` resolves at open: a finite, byte-range-seekable HTTP body (known `Content-Length` + `Accept-Ranges`) or an HLS playlist with `EXT-X-ENDLIST` is treated as on-demand; an open-ended response or a non-HTTP transport is live. On-demand throttles delivery and presents on a fixed 1× clock with a compressed read-ahead buffer, so VOD that arrives faster than real time no longer fast-forwards — and the caller doesn't have to flag it.

**3. Optional page-URL resolution.** `BasisMediaUrlRouter` is a small seam the player consults before loading a URL. `BasisMediaPlayerStreaming` routes its `StreamUrl` through it: a directly-playable URL (transport scheme, or an HTTP URL with a media extension) loads straight through; a page URL (no media extension) goes to an installed resolver. With no resolver installed, a page URL reports that the resolver package is needed instead of failing silently.

The new `com.basis.integration.ytdlp` package registers a yt-dlp-based resolver into that seam — turning a YouTube/Twitch page URL into the playable stream(s) (a split avc1+mp4a pair for VOD, a single HLS stream for live).

**4. Smooth live HLS.** Live HLS with **MPEG-TS segments** is paced by the engine's per-AU clock rather than a byte-rate token bucket, so variable-bitrate streams stay smooth through busy scenes while still presenting at — and converging to — the live edge. (fMP4/CMAF HLS — including Twitch low-latency — isn't rendered yet; see limitations.)

**5. Multiplayer-aware sync.** `BasisMediaPlayerNetworking` syncs the *page* URL, not the resolved CDN URL (which is per-client and expiring), so every client resolves it independently; direct stream URLs sync as before.

### Why a new seam (`BasisMediaUrlRouter`)

No existing player hook carries "resolve this URL before loading it," and the player package deliberately can't reference the integration package (removability — deleting the integration leaves the player intact). Dependency inversion via a registration seam is the only way to let the *existing* `StreamUrl` field steer page URLs without coupling the player to the resolver. The seam isn't an orphan extension point: its consumer (`BasisMediaPlayerStreaming.TryResolveAndLoad`) ships in this PR; the integration registers a second consumer. `IsDirectlyPlayable` lives on the player as the single classifier, and the resolver defers to it so the live-vs-resolve rule has one home.

### Scope / prerequisite

- **No resolver plugin ships in this PR.** The `com.basis.integration.ytdlp` integration is gated by asmdef `defineConstraints` (`BASIS_MEDIAPLAYER_EXISTS` + `YTDLP_EXISTS`) and declares no package dependency, so it **compiles to nothing** on any checkout without a provider plugin. The PR is safe to land as-is — the resolver stays dormant unless/until such a plugin is added (see "optional, dormant bolt-on" below).
- **Windows backend.** Split-stream and the auto-detect run on the Windows (Media Foundation) decode path. Other platform backends are follow-ups.

### Known limitations (documented in the package READMEs)

- Codec ceiling is **H.264 + AAC, ~1080p** — 4K YouTube is VP9/AV1-only, which the player can't decode, so resolution caps the chosen video at 1080p avc1.
- **Steering gap:** a direct HTTP stream with *no file extension* can't be told apart from a page URL; with a resolver installed it's sent to the resolver and fails rather than loading directly. Use a media extension or a transport scheme to avoid it.
- On-demand presents on a fixed internal buffer; `BufferMilliseconds`/`BufferMode` tune the *live* clock only.
- **fMP4/CMAF HLS not rendered:** HLS with fragmented-MP4/CMAF segments — including **Twitch low-latency** — loads a single frame then stalls with no audio. Standard **MPEG-TS-segment HLS works**; CMAF demux support is a follow-up.
- **Networked on-demand sync is start-together, not catch-up:** a client that joins after a VOD is already playing (or whose load lands later) starts from the beginning and won't match an already-running playhead until the next shared (re)load. The native backend exposes no absolute/forward seek to advance a client that's behind. Live streams converge to the live edge, and re-loading a URL restarts every client together.

## Demo

End-to-end: loading a range of sources through the BasisStereoMediaPlayer prefab in a Windows64 standalone build - confirming sync behaviour between two clients.

https://github.com/user-attachments/assets/29626748-1f80-4cdd-8d6f-9a6282a10262

## The yt-dlp integration is an optional, dormant bolt-on

The deliverable here is the **media-player work** — split-stream playback, live/on-demand auto-detect, live-HLS pacing, multiplayer page-URL sync, and a **generic** `BasisMediaUrlRouter` resolver seam. None of it depends on yt-dlp.

The `com.basis.integration.ytdlp` package is an **optional resolver that activates only when a provider plugin is present**. It's gated by asmdef `defineConstraints` (`BASIS_MEDIAPLAYER_EXISTS` + `YTDLP_EXISTS`) and declares no package dependency, so on any checkout without such a plugin the assembly is **not compiled at all** — no reference error, no package-resolution error, zero cost. With no resolver installed, a page URL simply reports that a resolver package is needed; everything else plays exactly as today.

**There is currently no committed provider plugin, and this PR doesn't need one.** The seam is generic, so any future resolver (yt-dlp-based or otherwise) can register against it without the player ever depending on it. Whether such a plugin is ever added — and how it would be vendored — is an independent question that does **not** gate this PR; the resolver stays dormant until a provider exists. Safe to merge as-is on the strength of the player work.

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

**What "Tested" covers (Windows):** *In the editor* — direct stream playback (RTSPT / fMP4 / TS / HLS) unchanged; a split avc1+mp4a VOD source playing in sync and paced (no fast-forward); auto-detect classifying a seekable HTTP body as on-demand and a live HLS stream as live; the `StreamUrl` field resolving real YouTube and Twitch links (and direct URLs bypassing the resolver); and the clear "resolver package needed" message when no resolver is installed. *In a built player* — live TS-segment HLS playing smoothly at 1080p60 under the per-AU pacing (standard Twitch / direct `.m3u8`; fMP4/CMAF is a documented limitation), a paced on-demand HLS VOD (`EXT-X-ENDLIST`) playing cleanly from the start and staying in sync across two clients, and the synced page URL resolving independently across two networked clients.

**Most checklist rows are N/A for this change** — it's load-path / native decode / integration code, with no per-frame gameplay work:
- No transform access, camera access, Addressables loads, or scene-wide discovery anywhere in the change.
- No new per-frame work, so nothing is added to `BasisEventDriver` and there are no hot paths — the "hot path" allocation/logging/collection rows don't apply. (The one cold-path delimiter array in `IsDirectlyPlayable` is hoisted to a `static readonly` field regardless.)
- No new `GetComponent`; `BasisMediaPlayerStreaming` uses the existing `TryGetComponent`.
- Jobification N/A — the work is socket I/O plus OS hardware decode on native threads, not parallelisable compute.
- Public fields (`Delivery`, `BasisMediaUrlRouter.Resolver`) are plain fields per house style; the resolver delegate is freely reassignable.

**Prerequisite:** the yt-dlp resolver only activates once the `com.yewnyx.ytdlp` plugin is present (separate PR). Until then this change is inert on that path and direct-URL playback is unaffected.

The yt-dlp embedding decision above is independent of the player work: split-stream, VOD pacing, live-HLS pacing, page-URL resolution and the multiplayer page-URL sync are all validated in the editor and in a built player (incl. live TS-segment HLS, a paced VOD HLS playing clean from the start, and two-client sync). The integration stays inert until a provider plugin lands.


towneh:
Ran another full validation pass on a fresh Windows64 standalone build with two networked clients, in two configurations: with the `com.basis.integration.ytdlp` resolver present, and with it removed from the project entirely.

**Direct transports** — RTSPT, HTTP fMP4, and HTTP MPEG-TS all play at the live edge and stay in sync across both clients, in both configurations.

**HLS** — live TS-segment HLS is smooth at 1080p60 under the per-AU pacing; a paced on-demand playlist (`EXT-X-ENDLIST`) plays cleanly from the start and stays in sync across both clients. fMP4/CMAF HLS still loads a single frame then stalls — that's the documented limitation and a separate follow-up.

**Multichannel** — LPCM 7.1 over MPEG-TS (`.m2ts`) plays via the decode-bypass path, paced correctly.

**Page URLs** — with the integration present, a YouTube watch page resolves to a split avc1+mp4a pair and plays in sync and paced. With the integration removed, a page URL just logs the "resolver package needed" warning and direct URLs keep playing — no crash, no package-resolution error, and the integration assembly isn't compiled at all. So the "compiles to nothing / safe to merge without a provider" claim holds up in practice.

One thing I found and split out: RTMP playback has a `createStream`/`play` sequencing bug (the stream plays fine in VLC but not in-client). It's a pre-existing transport issue unrelated to this PR, so I'm handling it as a separate fix rather than widening scope here — RTSPT remains the recommended low-latency PC path and works.

PR #892, opened 2026-06-19
Add BasisHighlight render feature

BoatFloater:
## Summary
I'd like to backport an outline/highlight feature I was working on in my own project to the Basis framework since I've been thinking about it recently. Let me know if you're interested. It is implemented as a custom render feature/pass, which there aren't currently any in the Basis framework.
Changes:
* **BasisHighlight**
  * A new framework-level render feature/pass to generate object outline highlights. It's built to be a relatively generic system which could be used by various sub-systems.
  * The render pass/feature are mostly vibe-coded, but it seems like the performance is good. It would be good to have a review by someone more experienced with URP. The changes to Basis internals are implemented by hand.
* **BasisPickupInteractable** 
  * Refactored to use BasisHighlight instead of its own highlighting system. The collider proxy system is re-used as a fallback if no `MeshRenderer` are found
* **PickupHighlightColor**
  * Renamed HighlightColor since it could be used for things other than pickup (if approved, all locales will need modification)
* **Examples**
  * Refactor the example button to use BasisHighlight in addition to its material recoloring 
  * Add a Highlight examples folder with the vertex offset shader to showcase `BasisHighlightOverride`
* **BasisPlayerInteract**
  * Not strictly related to the highlight feature, but I made the desktop line indicate from the dominant hand side of the screen and adjusted the destination point to be a bit more out of the way
* **Prefabs**
  * Re-enabled highlight on the personal mirror since it is an outline without fill; I suppose it's still a matter of personal preference
* **Content Police**
  * Allow `BasisHighlightOverride` component on Props, Scenes, and System

### Example
https://github.com/user-attachments/assets/7f5ae7e0-fb4e-4337-83b7-ab64767a2ccd

### BasisHighlightOverride
Setting a custom mask material so the outline can match vertex deform

https://github.com/user-attachments/assets/580717f9-f136-4cc1-85c6-945eed4d395f

### BasisPlayerInteract
Use dominant-hand and push nearest point out of the way
Screenshot 2026-06-19 155126
Screenshot 2026-06-19 155155

### BasisHighlightSettings
These are the settings I currently have set for both Desktop and Android, but they can be changed individually. I haven't tested on Android. The color is overridden by the runtime setting.
Screenshot 2026-06-19 155303

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes
Tested on PC in flat mode & with Valve Index


dooly123:
How does it go on Quest/Android? (performance wise)

dooly123:
happy to get it in overall, just need to make sure we are not impacting our weakest targets

BoatFloater:
@dooly123 I don't currently have a good way to test the performance on Quest; I will look into trying to set that up, but it would be good if someone who normally runs Basis on Quest could confirm.
Theoretically the shaders are written in a way that should run fine on the Quest GPU, and the default highlight settings could be lowered for Android if necessary.

BoatFloater:
I was able to get it up and running on Quest (2), in spite of all apparent efforts of the great team over at Meta.

I had to enable `requiresIntermediateTexture` for it to work on Quest. The glow passes also tanked performance, so I disabled glow and tuned some other settings for Android.

The new settings don't appear to affect my fps (read from the Basis menu) at all between 0 hovered objects (render pass is skipped) and 2 sustained hovered objects, though there appears to be a brief 1-4fps dip right when an object is first hovered; I'm not sure how that compares to developer though.

It would be good to get feedback from more people.

### New Android Settings
image

dooly123:
thought about this more, why dont we just disable it on quest / android that would be the safest way to handle any performance loss

dooly123:
(that way i can merge it)

BoatFloater:
I did a _slightly_ more scientific test and even with 125 objects highlighted which is way more than I suspect will usually be the case there was no noticeable fps difference on quest 2 with the new android settings. I did tweak them to be a bit sharper though.

I think it isn't necessary to disable it on Android. If it is, then more work would have to be done to wire a proxy renderer system into BasisHighlightManager so that anything that wants to highlight an object doesn't have to do double work.

Test looked like this (this screenshot is from the editor but I ran it on quest 2)
image

PR #914, opened 2026-07-01
Add local-player locomotion accessors (speeds / jump / gravity) + whitelist for sandboxed worlds

CompuGeniusCode:
**Problem:** Sandboxed world scripts (Cilbox) can't read or set the local player's locomotion — the character driver's movement speeds, jump height, and gravity aren't reachable from a world script, and the driver isn't whitelisted.

**Fix:** Add getter/setter methods on `BasisLocalPlayer` that proxy to `LocalCharacterDriver` — `Get/SetMinimumMovementSpeed`, `Get/SetDefaultMovementSpeed`, `Get/SetMaximumMovementSpeed`, `Get/SetJumpHeight`, `Get/SetGravityValue` — and whitelist them, plus the existing `Immobilize`, for scene scripts in `CilboxSceneBasis`.


yewnyx:
Note that multiple worlds can be loaded concurrently

ZevyaDev:
> Note that multiple worlds can be loaded concurrently
@yewnyx 

That's a fair point, however my PR matches the existing whitelisted player mutators (`Teleport`, `Respawn` and `Immobilize`), which all mutate the single shared `BasisLocalPlayer`. There's no per-world scoping primitive in the framework currently, so these are consistent with the rest of the exposed player API.
If Basis wants sandboxed player mutations scoped per world, that's a framework-wide concern that needs to be solved uniformly.

yewnyx:
Very good point. 

@dooly123 ^

Issue #908, opened 2026-06-28
Please add auto hold (check box) to vr controllers,


Crimsom13:
### Is your feature request related to a problem?

just tested on anewer build, much like desktop has auto hold we could use it for vr as well. Here's everything you need to hand to the Basis creator. The VR-hold feature is entirely in one Basis package (com.basis.framework/Interactions), and the old (un-reverted) copies are in your HubOne project.

The old scripts (file paths)

C:\Claude Projects\HubOne API Upgrade\Basis (HubOne Client)\Basis\Basis\Packages\com.basis.framework\Interactions\BasisInteractableObject.cs
C:\Claude Projects\HubOne API Upgrade\Basis (HubOne Client)\Basis\Basis\Packages\com.basis.framework\Interactions\BasisPlayerInteract.cs

What the change does

It makes auto-hold (sticky pickup) work in VR, not just desktop. Stock Basis only keeps an object held-after-release on desktop; this adds an opt-in VR mode plus a clean way to drop it. It's fully additive and backward-compatible (the new enum value is appended, so existing Yes/No serialized values are unchanged).

1) BasisInteractableObject.cs — three additions

a) New enum value (in BasisAutoHold, appended last):
/// Like Yes, but auto-hold also applies in VR — the object stays held after the grip
/// is released, until the dedicated drop button (see IsHoldDropTriggered).
/// Appended last so existing serialized Yes/No (0/1) values are unchanged.
YesIncludingVR

b) IsAutoHoldActive() — the device-aware check:
public bool IsAutoHoldActive()
{
    switch (AutoHold)
    {
        case BasisAutoHold.YesIncludingVR: return true;                                  // everywhere
        case BasisAutoHold.Yes:            return Device_Management.BasisDeviceManagement.IsUserInDesktop(); // desktop only (historical)
        default:                           return false;
    }
}

c) IsHoldDropTriggered() — overridable VR release hook:
/// Base returns true. Override for objects with specific hold/drop behavior.
public virtual bool IsHoldDropTriggered(BasisInput input)
{
    return true;
}

2) BasisPlayerInteract.cs — use those instead of the desktop-only gate

The grab/release logic was changed to call the new methods. The key line:

// BEFORE (stock — VR can never auto-hold):
bool autoHold = BasisDeviceManagement.IsUserInDesktop() && interactInput.lastTarget.AutoHold == BasisAutoHold.Yes;

// AFTER (the edit):
[...]

PR #917, opened 2026-07-02
Auto-hold: DesktopOnly/Everywhere modes + per-player VR opt-out

CompuGeniusCode:
**What:** Builds on this branch's VR auto-hold. Replaces `BasisAutoHold` `Yes`/`No` with `None`/`DesktopOnly`/`Everywhere`, so a world can auto-hold on desktop only or everywhere. `IsAutoHoldActive()` now resolves per-device.

**Opt-out:** New `DisableVRAutoHold` player setting ("Disable VR Auto-Hold") makes `IsAutoHoldActive()` return false in VR, overriding a world's per-grabbable value — the player wins over the world. Desktop is unaffected.

**Drop persistence:** Direct grab now requires a fresh grip press instead of a merely-held grip, so a pickup dropped by the world (or by the auto-hold toggle) stays dropped until the player intentionally grabs again — the still-held grip no longer instantly re-grabs it. Reuses the existing `wasGripDown` edge.

**Compat:** `DesktopOnly=0` (old `Yes`), `None=1` (old `No`), `Everywhere=2`; `CueGrip` updated to the new names.


ZevyaDev:
Closes https://github.com/BasisVR/Basis/issues/908

PR #927, opened 2026-07-09
feat(mediaplayer): HLS and LPCM (m2ts) playback on Android

towneh:
> **Stacked on #926** — review the last four commits (one per concern). Re-raised from BasisVR/BasisMediaPlayer#3 as part of the monorepo collapse, and now also carries the Android LPCM lane (folded in from #928 — both features gate on the same Quest pass from a single build, so one PR keeps the tested artifact and the merged artifact identical).

## Summary

Two Android audio/streaming gaps closed, one `.so`:

**HLS playback.** The HLS source (`protocol/basis_hls.c`) is portable C and already ships in the Android build — the only gate was `run_hls` constructing its HTTP provider under `#if defined(_WIN32)`. This wires the existing JNI `HttpsURLConnection` byte source in as the Android provider (same code path the push-model https fallback already uses: TLS, redirects and JVM thread-attach handled there), with a 60 s read timeout to sit out LL-HLS blocking playlist reloads. `.m3u8` URLs are handed to the HLS source ahead of the `AMediaExtractor` attempt — the extractor can't stitch segments, so trying it first only added a rejected network round-trip. Delivery semantics carry over unchanged from the portable core: live playlists present at the live edge with AU-level delivery pacing, ENDLIST playlists auto-detect as VOD (TS-segment VOD additionally picks up #926's seek/duration support wherever the portable layer provides it).

**LPCM (m2ts).** Brings the full-7.1 audio carriage to Android the way Windows has it (AAC decoders cap at 5.1). The demux side was already portable — the TS demuxer recognises Blu-ray HDMV LPCM (`stream_type 0x80`) and delivers header-stripped big-endian frames through the shared sink; Android received them and had nowhere to put them. On the LPCM codec no MediaCodec is created: `submit_audio` converts big-endian 16/24-bit Blu-ray-order PCM to interleaved WAVE-order float (the same channel-assignment remap tables as the Windows lane, which match ffmpeg's pcm_bluray remap and were verified by ear against a 7.1 channel-marker stream) straight into the PCM ring, whole frames only so the ring keeps its channel-phase alignment. 48 kHz / 16- or 24-bit, matching what the demuxer announces. One steering fix rides along: `AMediaExtractor` accepts `.m2ts` URLs but doesn't surface HDMV LPCM — it would play the video with the audio silently missing — so m2ts stays on the portable TS demuxer + bypass path.

No behaviour change on Windows for any of it (the extractor attempt is a no-op there and `run_hls` keeps its WinHTTP provider). Rebuilt `libbasis_media_native.so` (arm64-v8a); the Windows DLL is untouched.

## Status — draft pending device validation

Code-complete, **not yet Quest-tested** — one headset pass covers both features:

1. **HLS live**: TS-segment live HLS from the test origin (mediamtx, `mpegts` variant).
2. **HLS VOD + TLS**: a public TS-segment HLS VOD stream (`EXT-X-ENDLIST`) — also exercises redirects through the JNI fetch.
3. **LPCM 7.1**: `yuma_lpcm71.m2ts` (16-bit) — channel-isolation listening pass, same method as the 5.1 AAC validation: all 8 lanes audible, WAVE order correct.
4. **LPCM 5.1**: `blits_lpcm51.m2ts` (24-bit) — exercises the 24-bit conversion path.
5. **Regression**: stereo/5.1 AAC intact on the same build (the audio configure path gained a branch; the TS→AAC push path was already validated over RTSP on Quest).

Note for testers: `http://` origins need a build with cleartext traffic allowed (Android policy); `https://` and `rtspt://` origins need nothing special. The m2ts assets can be served over `rtspt://` (no extractor involvement) or `http://` (exercises the new m2ts steer).

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

Native C only, Android-guarded — no managed/per-frame code, so the driver/allocation/jobification items are N/A (the HLS work is network I/O on the existing threads; the LPCM conversion runs on the demux thread with a reused scratch buffer whose only allocation is one-time growth). "Tested" reflects the Windows no-behaviour-change leg (the guarded code compiles for both targets; Windows runs exactly as before — validated in-editor on the stacked branch, and the Windows LPCM lane validated this exact conversion and remap in its 7.1 listening pass when it shipped) — the Android device pass is precisely what the draft status above holds for. Kept as **draft until the Quest pass** listed under Status.


towneh:
Superseding this with #935 — the same commits (HLS, LPCM bypass, m2ts steering, RIST arm64, the AAC input-size fix) carried forward on top of the media-time sync work from #933, plus the on-device fixes from the Quest validation pass. Closing to keep review in one place.

PR #935, opened 2026-07-11
fix(mediaplayer): Android platform parity — HLS, LPCM 7.1, RIST, and the media-time sync model

towneh:
## Summary

Brings the Android backend to platform parity with the Windows media player and ports the media-time sync model to it. This supersedes #927, whose commits it carries; closing that PR in favour of this one.

**Playback surface (the former #927 scope):**
- **HLS on Android** — playlist handoff to the shared HLS source, riding the JNI HTTPS byte source.
- **LPCM 7.1 decoder bypass** — Blu-ray HDMV LPCM (m2ts) decoded directly to float PCM, bypassing MediaCodec; m2ts is steered past the OS extractor, which doesn't surface the LPCM track.
- **RIST enabled on arm64** (librist static, same opt-in build as Windows).
- AAC input buffers sized for multichannel frames (5.1 frames overflowed the codec's default input buffer and were fed truncated).

**The sync model (ported from the Windows implementation in #933):**
- Present clock v2: wall-rate advance with a rate-capped slew, clamped at the decode edge + buffer, ring-span-scaled paced resync, and a half-render-tick due-check lookahead. The first-generation clock's forced-present deadlock guard is gone — the edge clamp makes a stalled clock impossible.
- Decoded frames are held in a PTS-tagged AImage ring and presented when due, replacing acquire-latest presentation.
- The PCM ring serves by media time against the same clock, biased by the sink's reported output latency — banked mux lead no longer becomes audio latency, and just-in-time delivery banks a cushion instead of starving.
- Audio-first live start: when video is configured but not yet decodable (a mid-GOP join waiting for a keyframe), the clock bootstraps from the audio edge and video joins it in sync at the first frame — the VLC/ExoPlayer model.

**New fixes found in on-device validation:**
- **VOD delivery detection on Android** — the live-vs-VOD auto-detect was Windows-only, so a finite HTTP body classified as live and played at delivery speed (synchronised fast-forward). The JNI byte source now runs the same `Range: bytes=0-` probe as the WinHTTP source; the paced clock, delivery pacing and read-ahead all engage on Android.
- **`.m3u8` matched case-insensitively** in the HTTP dispatch, matching the sibling extension guards.
- **Media Players panel content clipped to its viewport** — the shared scroll-view prefab ships an unmasked, zero-anchored viewport, so panel content taller than the panel drew past its bounds; the panel now stretches and masks the viewport (standard scroll-view construction). Applies on all platforms.
- **JNI byte-source hardening** — anomalous zero-byte reads (outside the Java `InputStream` contract) are absorbed inside the source with a bounded retry, so they can neither masquerade as EOF nor spin the demux thread.

Rebased onto `developer` after #933 merged, so the diff is purely the Android work. The rebuilt arm64 `.so` (RIST-enabled) is committed, and the Windows DLL is refreshed to carry this branch's shared-dispatch changes.

## Required checks

## Testing details

- [ ] Windows
- [ ] Linux
- [x] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

### Verification (Quest Pro, Horizon OS / Android 14, VR)

| Lane | Scenario | Result |
| --- | --- | --- |
| RIST, AAC 5.1 | plain | in sync, clean |
| RTSPT live, AAC 5.1 | normal join | in sync; rare single-block click at the test feed's loop seam |
| RTSPT live, AAC 5.1, ~10s-GOP source | joined mid-GOP | audio starts immediately, video joins in sync at the first keyframe |
| HTTPS live, HTTP-TS AAC 5.1 | over TLS | in sync, clean |
| HTTPS HLS VOD (TS segments) | over TLS | plays (exercises the Android HLS path end-to-end) |
| HTTPS progressive VOD, LPCM 7.1 (m2ts) | over TLS | paced at normal speed, in sync (exercises VOD detection + LPCM bypass + prime) |

- All measurements via the media player's diagnostics CSV on-device: audio queue steady, clock-gated trims ≈ 0 in steady state on every lane, A/V in sync by ear throughout.
- Known minor, tracked separately: on the highest-bitrate VOD asset (27.8 Mbps over headset Wi-Fi) the video frame bank occasionally drains below the hold, showing as a single held frame — being A/B'd against a lighter asset to separate bandwidth marginality from pacing behaviour before any code change.

### Checklist N/A context

- The change is native C (decoder/present/demux/JNI) plus two scoped C# touches: the panel provider's viewport containment (runs once at panel build; `TryGetComponent` first, `AddComponent` only when absent) and the splitter/tap rate-conversion parameter. No transforms, no asset loads, no camera access, no scene discovery.
- Jobification isn't applicable: the work runs on decode/render/audio threads in the native plugin, not on the main thread.
- No new logging on hot paths; the native engine's existing rate-limited `basis_media` log tags are unchanged.


towneh:
Opted to go with a fix prefix here rather than feat. Whilst both are true this fixes more than it introduces.

PR #947, opened 2026-07-13
feat(mediaplayer): Android byte-source reseek parity (JNI ranged refetch + abort)

towneh:
## Summary

> **Verified on Quest (Quest Pro, Android 14).** The one runtime behaviour that needed a device — Android's `HttpURLConnection.disconnect()` unblocking a thread parked in `read()` — is confirmed: the MP4 byte-source seek repositions cleanly both directions, no decoder errors, playback resumes at the target.

The native VOD reseek path — absolute seek, and (now that #946 has landed) the trailing-`moov` range-fetch — was Windows-only because only the WinHTTP source could do a ranged refetch and abort an in-flight read. The Android JNI HTTP source detected seekability (the `206` probe used to pace VOD) but had no way to reposition, so the demuxer got a null reseek hook there and every range-fetch path fell back to non-seekable behaviour.

This adds the two missing primitives to the JNI source:

- **`basis_jni_https_reseek(offset)`** — opens a fresh `HttpsURLConnection` GET with `Range: bytes=offset-`, checks for a `206`, swaps in the new stream. The connect sequence is factored into a shared helper so `open` and `reseek` build the request identically.
- **`basis_jni_https_abort()`** — disconnects the connection so a read parked in `InputStream.read()` on the reader thread throws and returns at once (the counterpart to closing the WinHTTP request handle).
- **`basis_jni_https_can_reseek()`** — the `206`-proven gate for wiring the hook.

`http_reseek` in the core is generalised to take the abort/reseek primitives as function pointers, so the park-the-reader / flush-the-ring choreography lives in one place and the Android wiring branch just supplies the JNI versions. The Windows path is unchanged in behaviour (same calls, through the pointers).

Self-contained in the native layer — no C# or public API change.

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [x] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

**Device verification (Quest Pro, Android 14):** the abort's dependency — `HttpURLConnection.disconnect()` unblocking a thread parked in `InputStream.read()` — is confirmed on device. MP4 byte-source seek against a `https://` range/`206` VOD (Big Buck Bunny over the JNI-HTTPS path) repositions cleanly forward and backward, with no decoder error in `adb logcat` and playback resuming at the target.

**What *is* verified:**
- Android arm64 `.so` builds clean (RIST-enabled), no warnings from the changed files.
- The shared core change (the function-pointer `http_reseek`) also compiles into the Windows DLL, so I regression-tested it there: drove a VOD seek over a `206` server — seek returns `0`, playback continues past the target (frames keep decoding, position advances), no error. A null/wrong pointer would crash, so this confirms the indirection is correct and the Windows path is behaviourally unchanged.

**Required checks marked N/A collectively:** this is a native-C change (`android/basis_jni_https.c` + `basis_media_core.c`) plus rebuilt plugin binaries. No C#/MonoBehaviour, Unity API, per-frame, transform, component, camera, Addressables, or managed-logging surface is touched — so every C#/Unity-side check is N/A. Native logging stays on the existing `__android_log_print` path the rest of the Android layer uses (`BasisDebug` is C#-only).

**Redirect handling:** the status gate rejects `3xx`, not just `4xx`/`5xx`. `setInstanceFollowRedirects` follows same-protocol redirects transparently, so a surviving `3xx` is a cross-protocol `http`↔`https` hop that `HttpURLConnection` won't follow — `getInputStream()` would return the redirect page, not the media. Failing cleanly beats demuxing that. Note the resulting asymmetry: WinHTTP follows `307`/`308` and same-protocol upgrades itself, so those fail cleanly on Android where they'd play on Windows; hand-rolling cross-protocol redirect follow (with the same downgrade guard WinHTTP applies) would be its own change.

**Platform scope:** the reseek hook is wired on both the Windows and Android byte sources now. Both plugin binaries rebuilt. Pairs with the outstanding Android VOD-seek work — this is the byte-source half of it.


towneh:
Verified on device — Quest Pro (Android 14).

The Android MP4 byte-source seek repositions cleanly both directions against a range/206 https VOD (Big Buck Bunny over the JNI-HTTPS path): no decoder errors in logcat, playback resumes at the target. Also sanity-checked live TS, HLS load, and the security refusals (private-IP + `file://`) — all as expected.

Testing did surface a few pre-existing issues in adjacent code this PR doesn't touch — a coded-height pad showing as a strip on non-16-aligned heights on Android, an HLS-VOD seek quirk, and a backward-seek audio lag on high-bitrate content. Flagging them here so they're not read as regressions from this change; I'm tracking them separately.

PR #952, opened 2026-07-13
MediaPlayer: Seek for integrated fMP4 file

cathiecode:
## Summary

This pull request enables video seeking on videos from certain websites that stream Integrated fMP4 files.

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [ ] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

### Validation
Testing was conducted in the editor on Windows 11.
Integrated fMP4 VOD file: https://zipline.space.superneko.net/raw/bbb_sunflower_1080p_30fps_normal_idfmp4.mp4

The test file was encoded using the following options:

```powershell
ffmpeg -i input.mp4 `
    -map 0:v:0 -map 0:a:0? `
    -c:v libx264 -profile:v high -level:v 4.0 -pix_fmt yuv420p -r 30 `
    -c:a aac -b:a 128k -ar 48000 -ac 2 `
    -video_track_timescale 15360 -movie_timescale 15360 `
    -x264-params "keyint=210:min-keyint=1:scenecut=40:open-gop=0" `
    -movflags +empty_moov+dash+frag_keyframe+default_base_moof+global_sidx+skip_trailer `
    -brand dash `
    -use_editlist 1 `
    out.mp4
```

* Integrated fMP4 VOD: Pausing and then resuming playback causes a small jump of approximately one second. This matches the behavior of progressive MP4 playback, so there is no regression. Player on Android also works.

* HLS VOD: The video is reported as seekable, but playback becomes stuck after actually seeking. This matches the original behavior, so there is no regression, although it is probably a bug.

* RIST: Validated successfully.

* Split-stream playback: Tested with a video podcast resolved by `dlp-native`. For some reason, the audio briefly cuts out approximately one second after audio/video synchronization completes. After this interruption, playback continues without further issues.

### Limitations
- Seek requests are snapped to the start of the earliest fragment containing that timestamp.
- Currently, this code assumes that a file has exactly one sidx box. at least, I have confirmed that a 1.5-hour video on the website is delivered in this format.
- Seeking to SAP-type-0 (unspecified) fragment may cause visual artifact, but I skipped this check this intentionally. I couldn't find any resource that describes how I encode/demux the test material using ffmpeg.

### Native libraries
Base native libraries are built with this patch, based on 59b6e043146a9e8b195e77ab5d6576ec8e9f88c4 librist enabled.


towneh:
Confirming this one's known — we hit the same on device (Quest, Mux `x36xhzz` HLS VOD) and had it tracked separately, so agreed: pre-existing, not a regression from this PR.

Light version of what we've seen so far: the segment jump itself works (the VOD seek index rebuilds the fetch queue at the target), but the delivery pace clock doesn't re-anchor on an HLS seek the way a byte-source seek does — so after the jump, playback runs against a stale pacing anchor and stalls. It's on our list; still digging, no fix pinned down yet.

towneh:
Follow-up: the HLS-TS VOD seek issue mentioned on this PR (forward stall / backward flood after seeking) is now fixed in #956 — the segment-source reposition re-anchors the delivery pace clock at the flushed boundary.

towneh:
Closing the loop: this is fixed in #956 (merged). One correction to my note above: a re-anchor did fire on HLS seeks, it just lost a race to access units already in flight from before the seek, so the pace anchor could re-base on stale timestamps. #956 moves the flush and re-anchor onto the demux thread at the flushed boundary, the same way the byte-source seek already worked, and both seek directions now behave on the Mux fixture (Windows editor and Quest).

PR #973, opened 2026-07-19
feat(constraints): Basis constraint component, solver, inspector and tests

dooly123:
The `Constraints/` package landed a data layer (`BasisConstraintSlot`, `BasisConstraintMath`) with nothing using it — zero references anywhere in the repo. This adds the component layer that replaces the six `UnityEngine.Animations` constraints, the solver behind it, a Basis-styled inspector, and test coverage.

## Design decisions worth a look

**One component, not six.** `BasisConstraint` has a `Kind` dropdown rather than shipping `BasisPositionConstraint`/`BasisRotationConstraint`/etc. Reasons: it matches `BasisConstraintSlot.Kind`, it keeps a single batching path, and it avoids shadowing the existing `Basis.Scripts.BasisSdk.Interactions.BasisParentConstraint`, which is serialized **by name** in scenes (`InteractablesScene.unity` and others). Happy to switch to six thin subclasses if you'd rather have 1:1 drop-in parity with Unity's components.

**One blending contract for every kind**, since the six Unity constraints each behave slightly differently and that is a common source of confusion:
- `Active == 0`, or no source above zero weight → exact no-op.
- `Weight == 0` → exact no-op, never a snap to rest. (Blending from rest would make the result discontinuous at 0; there's a regression test for this.)
- Axes excluded by a mask hold the **live** pose when `Locked`, or the **rest** pose when unlocked. That gives both `Locked` and the `*AtRest` fields a real, testable meaning.

**Slots are ordered shallowest-first** by hierarchy depth, so a constraint following a transform that is itself constrained reads the solved pose rather than a stale one.

## Editor UI

`BasisConstraintInspector` follows the network-sync house style exactly: UI Toolkit only, `BasisSyncInspectorUI` `Header`/`Card`/`Described`/`ValidationContainer`, `BasisDocInspector_UI` base, the `Header → Validation → Cards → Bind → API foldout` ordering, tooltips on every control, and a live readout on a 300 ms schedule.

Cards show only for the kinds that actually read them (a Position constraint isn't padded out with aim vectors), and the live "Effective blend" readout shows how weights actually normalise — a lone source at 0.25 still fully drives the target, which surprises people. Validation flags self-referencing sources, ancestor sources on Parent, fully-masked channels, and a `World Up` mode with no up object assigned.

## Tests

`Basis.Framework.Constraints.Tests` — 2 suites, ~50 cases:
- **Solver**: the no-op contract, weight normalisation and proportional blending, slot-weight lerp, source offsets applied in source space, axis masks locked vs unlocked, per-kind channel isolation (a Position constraint must not touch rotation/scale), quaternion hemisphere handling (blending `q` with `-q` returns `q`), Aim/LookAt aiming invariants, roll, world-up modes, and batch ordering for chained constraints.
- **Builder**: component→slot projection, weight clamping, degrees→radians conversion, disabled-GameObject handling, transform de-duplication, disjoint source ranges, depth ordering, and an end-to-end build→read→solve→write pass.

Aim tests assert the invariant ("the aim axis ends up pointing at the source") rather than hardcoded quaternions, so they don't encode a twist convention.

## Verification

Type-checks cleanly (0 errors, 0 warnings) against Unity 6000.5.4f1 assemblies and the compiled `Basis Framework` / `Basis Framework Editor` / `BasisEditor` DLLs. Test *execution* results are noted in a follow-up comment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dooly123:
## Test run

Executed the suite in Unity 6000.5.4f1 batchmode against this branch (EditMode, filter `Basis.Tests.Constraints`):

```
test-run  testcasecount="61"  result="Passed"  total="61"  passed="61"  failed="0"
  BasisConstraintSolverTests    34/34
  BasisConstraintBuilderTests   27/27
```

Whole-project compile was clean — zero `error CS` across all 2337 build nodes, including `Basis.Framework.Constraints.Tests.dll`.

Run twice to confirm; identical 61-case list both times.

### One bug found and fixed while testing

`CaptureRest` stored `transform.localPosition` / `localEulerAngles` / `localScale`, but `BasisConstraintSolver` resolves entirely in **world** space. For any transform below the scene root the unlocked-axis rest fallback would land in the wrong place, silently — a child at local `(1,0,0)` under a parent at `(10,0,0)` recorded a rest of `1` instead of `11`. Now captures world pose, with `CaptureRest_OnAChild_CapturesWorldSpaceNotLocal` as a regression test (it fails against the previous implementation).

### Notes for review

- `BasisConstraintData.cs` and `BasisConstraintMath.cs` were untracked working-tree files; they're included here since nothing compiles without them.
- Test *execution* is verified; what is **not** verified is how the inspector actually looks rendered in the Unity Inspector window. Worth a visual pass — it's built from the same `BasisSyncInspectorUI` primitives as the sync inspectors, so it should sit consistently alongside them, but I couldn't eyeball it headlessly.

Toys0125:
I want to comment that you should also test that vixxy works toggle the component and having multiple on the same object.

PR #996, opened 2026-08-06
fix(mediaplayer): harden the native core against hostile stream sources

towneh:
## Summary

A security review of the Media Player's native core turned up 44 findings. This branch closes 41 of them. Five further findings came out of the work itself, and four of those are closed here as well. It is native C/C++ only: no managed code changes.

**Why this matters.** Media URLs are chosen by whoever drives the player, so this code parses bytes from a host the local user did not pick, in-process. The review found one critical and seven high-severity issues in that path, all reachable by being in an instance with someone who can set a URL.

**Read the behaviour-changes section before the fix list.** Most of this is hardening that no user will notice, but three changes alter what the player does, and those are the parts worth your attention rather than the bounds checks.

**Build hardening.** The native plugin carried nothing but warning levels, leaving every mitigation to toolchain defaults, and those defaults differ sharply between MSVC and clang. Windows x64 now builds with Control Flow Guard and CET shadow-stack compatibility (`dumpbin` reports both, plus a 394-entry guard function table, against none of the three on the previously shipped binary). The GCC/Clang branch probes each flag rather than assuming it, since Apple clang, NDK clang and gcc each reject a different subset. The Linux binary has now been rebuilt through that path and checked on the artifact rather than at the compiler: RELRO, BIND_NOW, a non-executable stack, the stack protector and FORTIFY are all present, which had never been verified on a Linux output before.

**Lifetime and threading.** The critical finding is a use-after-free: `basis_media_close` freed the engine and decoder while the Unity audio thread could still be inside `basis_media_read_audio`. Closing it needed a liveness registry with per-engine audio slots, so close deregisters and waits out any in-flight pull before freeing. Alongside it: the video texture was handed to Unity as three unlocked loads racing the demux thread's rebuild, and both HTTP byte sources could have their handles swapped out from under an in-flight read. Teardown also aborts the byte source before joining its reader, which both platform sources supported and the portable one did not — so on the builds that use it, the join sat out the socket's full read timeout instead of returning promptly.

**Redirects.** Following redirects by hand meant the address policy only ever saw the entry URL, so a redirect walked straight past it into loopback, link-local or RFC1918 space. Every hop is now re-validated. Each backend resolves the `Location` with its own machinery (`UrlCombineW`, `java.net.URL(URL,String)`) rather than a second parser, which would be a differential-bypass surface. The loop also tracks the entry transport, because following redirects by hand drops WinHTTP's `DISALLOW_HTTPS_TO_HTTP` and the JRE equivalent, which would otherwise let media stream in the clear over an https to http hop.

**Bounds on remote input.** RTMP chunk reassembly, the RTSP header block and body, the portable HTTP header block, the RTP timestamp conversion, the SEI caption accumulators, the HLS playlist durations and the delivery pace gate all took attacker-declared values into arithmetic or allocation without a ceiling. Each is now bounded at the point the value enters.

**Tooling.** `tools/media-fuzz` gains a WAV target, since `basis_wav.c` walks RIFF chunk lengths out of the same untrusted bytes and had no coverage; 24M runs under ASan/UBSan found nothing. The RTMP allocation cap is now gated, which it was not before: a plain replay proves nothing about a *bound*, so that repro runs under `-malloc_limit_mb=8` and was verified to fail with the cap removed and pass with it in place. `TESTING.md` is explicit that this gates the per-message cap only — `-malloc_limit_mb` bounds a single allocation, and the per-session total accumulates across many, so it cannot be gated the same way. The RTSP endless-header hang cannot be gated either, because the fuzz stub is finite, and the doc says so instead of implying CI covers it.

### Behaviour changes a reviewer should weigh

Three changes here alter what the player does rather than only what it refuses.

**1. The Android decode path is now the portable demuxer.** Android used the OS `AMediaExtractor` to demux MP4 and WebM. That extractor ran its own HTTP and followed its own redirects, so the entry host was checked and then it went wherever the origin pointed it, from the headset's position on the user's LAN. It no longer runs on this platform at all. The portable `basis_mp4`/`basis_webm` demuxers read every container through the JNI HTTP byte source — where the address policy is enforced on every hop — and feed MediaCodec through the same `submit_video`/`submit_audio` path Windows and the RTSP/RTMP leg already use. One demux path on every platform, hardware decode kept, ranged-refetch seeking kept, and no OS component doing I/O the policy can't see. It also refuses VP8, which the portable WebM demuxer does not carry and the extractor would have played. **This is the largest behavioural change in the branch and it has been verified on a Quest Pro (see the device pass below).**

**2. Linux gains five container formats.** The shipped Linux binary had been built from a partial source list and contained no WebM, Ogg, WAV, MP3 or HLS. Rebuilding it through CMake restores them: exported functions go from 137 to 228, with no symbol lost in the other direction. That is a functional gain arriving inside a security PR, and it wants testing on that footing rather than as a like-for-like refresh.

**3. Windows TLS policy.** The WinHTTP session now sets a TLS 1.2 floor and turns on certificate revocation checking. The floor fails the open, like the timeouts already set there. Revocation deliberately does **not**, because its failure mode is refusing a good stream when the revocation endpoint is unreachable rather than refusing a bad certificate, and a client behind a restrictive network should still play. If you would rather that were strict too, it is a one-line change.

Two smaller ones: RTSP requests are no longer issued once a stop has been signalled, with the post-PLAY `TEARDOWN` deliberately exempt so the server's session is still released (suppressing it would turn a stop/start into a failed restart on session-limited devices such as cameras); and the HTTP parsers are stricter about `Content-Length` and chunk sizes, so a malformed value now fails the open where it used to be taken as a numeric prefix.

### What this closes

The review classified one issue critical and seven high. This branch closes the critical and all seven:

| Finding | Severity | State |
| --- | --- | --- |
| Audio thread reads a freed engine | **Critical** | Fixed, reproduced then clean |
| Texture handed out as three unlocked loads | High | Fixed, 391 use-after-free to 0 |
| Redirect bypasses the address policy | High | Fixed, verified on both backends |
| Redirect leaves TLS behind | High | Fixed, verified on both backends |
| RTMP chunk reassembly unbounded | High | Fixed, fuzzer A/B |
| RTSP header reader unbounded and uninterruptible | High | Fixed, harness A/B |
| Unclamped `snprintf` return used as a write length | High | Fixed, leak demonstrated then absent |
| Unbounded read in the portable HTTP open | High | Fixed, hang reproduced then bounded |

The last two both live in `basis_http_open` and were closed together. That function is reached only on the `#else` arm of the byte-source dispatch — Windows takes WinHTTP, Android takes the JNI bridge — so it is live on the Linux binary this repo ships, for plain `http://`. The buffer it formatted requests into was 2048 bytes when the largest request a parsed URL can produce is about 2400, so it could not represent a legal request at all and every long URL truncated. `snprintf` answers with the length it *wanted*, and that value went straight to the socket write as a length. Against a 2000-character path the pre-fix build sends 2106 bytes of which the last 59 are not its own, and the origin receives two live stack addresses and several other pointers. That is the ASLR bypass the finding predicted, so it is worth being plain that this was a real disclosure primitive rather than a theoretical one.

All thirteen mediums are also closed. The ones worth naming: the reader-thread lifetime on both byte sources; `http_reseek` reallocating under the reader; `pthread_create`'s return discarded while the join flag was set regardless; `send` without `MSG_NOSIGNAL`, where a peer that half-closes could raise `SIGPIPE` and take the whole client down; the delivery pace gate subtracting two container timestamps directly in `int64`; the RTP timestamp conversion guarding its division and not its multiplication; the SEI accumulators, where signed overflow made the bounds check below them removable; the HLS durations converted with `atof` and cast to `long`; and the WinHTTP session created with no protocol floor and no revocation checking.

Twenty of the twenty-three lows are closed, including both PCM ring allocations, which had been deferred earlier and are now done: the initialisers report failure and the call moved ahead of anything needing unwinding, so a failure is a plain free rather than a teardown path of its own.

Reviewing the branch turned up further defects in its own new code, fixed here rather than left for later. The ones worth knowing about, because they were introduced by fixes in this PR: a reference-table use-after-free in the JNI source, where a quiesce used a snapshot after releasing the lock that protected it, reachable when the read-ahead reader and a reseek run concurrently. Also the RTSP status line read from memory never written for it, `NewGlobalRef` results installed unchecked, the audio-slot claim retrying unbounded on a deadline thread, and the Android log helper stripping userinfo at the first `@` rather than the last.

The playlist number parsing is worth its own note, because it took three attempts and the two failures are instructive. Replacing `atof` with `strtod` looked like the obvious fix and was not: `strtod` takes its radix character from `LC_NUMERIC`, so on a host process running a comma-radix locale it would have stopped at the decimal point and returned a truncated value that still passed the "did anything parse" check — silently reproducing the exact defect it was meant to remove. The hand-rolled replacement then tested its ceiling *after* multiplying, which is undefined once the accumulator is near the top and `long` is 32 bits, as it is on Windows. Both are fixed and both are covered by the test described below.

Five findings were raised after the original review, and four are closed here: the 21-second teardown stall inside an uncancellable WinHTTP open (bounded; the cancellation half is still open); the Android redirect bypass above (closed by routing that platform through the policy-checked byte source instead of the OS extractor); the absence of `SO_SNDTIMEO` on any socket, where a peer holding a zero window blocks the demux thread indefinitely for as long as it likes; and the write-side `EINTR` gap, where a signal arriving during a request abandoned a working session. The fifth is the transport question described under what is left open.

### What is left open

Three of the original 44, and one later finding. None is left through oversight.

The three are the DLL-planting findings — the `MFTEnumEx` flags and the `opus.dll` load. They are accurate, but that enumeration is what selects the VP9, AV1 and MP3 decoders, and the failure mode when it goes wrong is a silent software fallback that no amount of local testing surfaces. They want a codec-matrix device pass, or an explicit decision to accept the risk, not a flag change shipped alongside unrelated work.

The fourth is a transport question rather than a bug: the https-to-http refusal is scoped to a single request and its redirect chain, so an https HLS playlist may list `http://` segments and they are fetched in the clear. Closing it means refusing mixed-scheme playlists, which exist in the wild and play today, so it is a compatibility decision worth taking on its own rather than inside this branch.

### How each fix was proved

The bar here was demonstration rather than argument: reproduce the bug on the pre-fix build, then show it absent on the fixed one, using an instrument that can actually witness the failure. A race that is merely "not observed while clicking around" has not been tested, and several of these produce no visible symptom at all until they corrupt something.

| Fix | Instrument | Before | After |
| --- | --- | --- | --- |
| Audio thread vs close | concurrency stress harness | reproduced in 4 iterations | clean |
| Reader-thread lifetime | seek stress against a 133 MB source | ABA 2 per run, three runs | 0 per run |
| Texture handoff | D3D11 + D3D12 probe driving a real device through `UnityPluginLoad` | 391 use-after-free; D3D12 faulted outright (`0xC0000005`) | 0 on both |
| RTMP bounds | libFuzzer A/B | OOM at `malloc(16 MiB)` from 8.6 KB of input | clean, 20k runs |
| RTSP bounds | standalone endless-header server | ran to a 64 MiB ceiling | stopped at 600 bytes |
| HTTP header bounds | endless-header origin, A/B against pre-fix source | hung until the harness killed it | refused in 1.3 ms |
| HTTP request write | origin reporting what it received, 2000-char path | 59 bytes of stack, two live addresses | correct request tail |
| Redirect policy | 16 redirect fixtures, loopback listener as witness | 1 loopback connection per fixture | 0 |
| Playlist number parsing | 33-case unit test, run in the C locale and `de_DE.UTF-8`, compiled `-m32` under UBSan | `signed integer overflow: 999999999 * 10` | zero UBSan reports, identical results in both locales |

The HTTP header row needed a middle case to mean anything: a 150-header response is accepted by the pre-fix build and refused by the fixed one, while a normal 10-header response still opens on both. Without that, "the endless case stopped" could have been a socket error rather than the cap.

Getting those numbers honest took more work than the fixes did, and several false passes are worth recording so nobody repeats them. ASan cannot see the reader-thread bug, because a `HINTERNET` is an OS handle rather than heap. A "did the handle change" probe conflates the benign detach with the real reallocation. A small test fixture finishes downloading before the first seek, so nothing is ever in flight and the race cannot occur, which is why the seek harness uses a 133 MB source. The texture probe had to hand the plugin a real D3D11 device, because outside the editor `basis_gfx_get_d3d11_device()` returns NULL and a naive harness "passes" against code that never ran. And the HTTP harness initially reported every case failing in 0.0 ms, because its origins are on loopback and the address policy was correctly refusing them.

The parser test earns its row for the same reason. Its first version ran only at the machine's native word size, where `long` is 64 bits, and passed everything — including the input that overflows at the 32-bit width Windows actually ships. A test that cannot fail on the target platform is not evidence about it, so the parser cases now compile `-m32` under UBSan, and the test's copies of both parsers are checked byte-for-byte against the shipped ones before any result is believed.

One fix is weaker than the rest and should be read that way: the seek park handshake was corrected after review on the argument that a previous reposition's acknowledgement could satisfy the current one. The vulnerable state is provably reached, seven times per run, but no reposition ever landed inside it, so that one is reasoned from construction rather than reproduced.

The Android paths were exercised on a Quest Pro rather than compile-checked, because they are genuinely different code: `pthread_mutex` and condvars where Windows uses SRWLOCK, and JNI where Windows uses WinHTTP. A full pass covered every transport and codec lane, teardown under live audio, pause/resume/stop/replay, a seek stress run including stop mid-seek, the redirect matrix, RTSP with a stop mid-stream, and the negative lanes (RFC1918, `file://` and loopback each refused with a reason, not a timeout). No crashes or ANRs on any lane.

After the decode path moved to the portable demuxer, the MP4/WebM codecs that had only ever run through the old OS extractor were re-verified on device: H.264, HEVC and the fragmented / `sidx` containers play smoothly at 1× with working seeks (decode rate matches present rate, clock at 0.99×), AV1 refuses cleanly with no crash (Quest Pro has no hardware AV1 decoder), and VP8 refuses. The rest of the MP4/WebM matrix — VP9-in-WebM, AAC 5.1, Opus — had already run through the portable path.

Two device results are worth calling out because they are easy to fake and easy to get wrong:

- Every redirect refusal was checked against the native log to confirm it came from our own guard (`refusing blocked host 127.0.0.1`, `refusing https -> http downgrade on a redirect`) rather than from a dead port. Worth noting that Android's cleartext policy was never a backstop here: this app sets `usesCleartextTraffic="true"` and a build processor enforces it, so a plaintext redirect target would have been allowed by the platform. The refusals came from our code because nothing else would have refused them.
- The reposition path throws a Java exception into the reader on every seek by design, and all eight were confirmed cleared. A single missed clear aborts the process through CheckJNI on the next call, so that is a load-bearing detail rather than tidiness.

## Required checks

## Testing details

- [x] Windows
- [ ] Linux
- [x] Android
- [ ] iOS
- [ ] macOS

- [ ] Avatar swapping

## Notes

**On the required checks.** This branch contains no managed code. It is native C/C++, CMake, the fuzz harness and its docs, `TESTING.md`, one CI workflow file and the three rebuilt binaries, so the Unity-shaped boxes (transform access, Addressables, `GetComponent`, `BasisEventDriver`, jobification, properties, camera driver, `BasisDebug`, scene discovery) have nothing to apply to and are ticked as N/A per the template's own instruction. The four that genuinely apply were treated as real: no allocations in hot paths, no debugging in hot paths, hot-path collection access, and Tested.

The one place worth a reviewer's eye on hot-path cost is the texture handoff. It now takes a lock and retains a reference on every `basis_decoder_get_texture` call, which is per-frame. The lock is uncontended in the common case and the retain is a single `IUnknown::AddRef`, and it measured clean over 335 D3D11 and 91 D3D12 rebuild cycles, but it is a real change to a per-frame path and I would rather flag it than have it found in review.

**Tested on** Windows (Editor, D3D11 and D3D12) and Android arm64 on a **Quest Pro**. The change does not touch player, XR, input, avatar or networking code, so hot-switching, avatar swapping and server swapping are unaffected.

**Linux is left unticked deliberately.** The binary is built, its mitigations verified on the artifact, and its exported API diffed against the one it replaces — but nobody has started a client with it. Building is not testing, and it would be the wrong box to tick given it also enables five container formats that binary never had.

**Which binaries carry these fixes.**

| Binary | State |
| --- | --- |
| `Plugins/Windows/x86_64/basis_media_native.dll` | rebuilt |
| `Plugins/Android/arm64-v8a/libbasis_media_native.so` | rebuilt |
| `Plugins/Linux/x86_64/libbasis_media_native.so` | rebuilt (was pre-fix, and predated the review) |
| `Plugins/Windows/ARM64/basis_media_native.dll` | rebuilt |

Linux was the one that mattered most, because it is the only shipped target where `basis_http.c` is the live byte source, so it carried both of the high-severity findings in that file on top of the critical. It is now current, after one more fix late in the branch: the decoder interface's seek entry point had no stub implementation, and because ELF linking tolerates undefined symbols in a shared object, the earlier rebuild shipped with `basis_decoder_seek` unresolved. A first seek would have faulted at symbol bind. The stub now provides the no-op and `nm -D` on the committed binary resolves clean.

Windows ARM64 is now current too. Its CMake path had been broken twice over: `run_http_like` selects the WinHTTP byte source on `_WIN32` regardless of which decoder backend is in play, but the stub branch never compiled `windows/basis_win_http.cpp`, so the link failed on unresolved symbols — repaired earlier in the branch. The second break was the missing `basis_decoder_seek` stub above, which MSVC caught as LNK2019 the moment the target was first built through CMake with a real ARM64 toolset. That link failing is the same defect the Linux `.so` was silently carrying; MSVC just refuses at link time what ELF defers to runtime. The committed DLL is a genuine ARM64 artifact (PE machine `0xAA64`). It remains the stub backend — it loads, exercises the portable protocol and demux core, and produces no frames — and it has not been run on a Windows-on-ARM device.

**One further known gap, stated so this does not imply coverage it lacks.**

BTI is not enforced in the committed Android `.so`. The codegen is present (132 `bti`, 187 `paciasp`) and our own object files carry the AArch64 property note, but lld only emits it on the output when *every* input object carries it, and the prebuilt `librist.a` this binary was linked against predates the flag. `build-librist.sh` in this PR does add `-mbranch-protection=standard`, so rebuilding librist with the committed script resolves it. That archive is not tracked in the repo, so the rebuild has not happened here. The `android-arm64` CI job does build it with those flags and publishes it as an artifact, so dropping that into `third_party/` and relinking is the shortest route if you would rather this were closed before merge. PAC is unaffected, as it needs no loader opt-in.

Not exercised on device: RTMP, because the test server binds 1935 to loopback, and the RTSP bounds half, which needs a hostile server on a reachable port. Both were fuzzer-verified on Windows and the Android binary compiles the same code.

**Two deleted build scripts, and where that left things.** The hardening commit removes `build-win-arm64.bat` and `build-linux-from-windows.bat`. Both were Zig cross-builders that invoked `zig c++` directly and bypassed CMake, so they produced binaries carrying none of the mitigations this PR adds — and, as it turns out, missing whole source files: the ARM64 script's hand-written list omits seven of the demuxers.

Removing them looked risky at the time, because they were the build scripts for the two stale targets. In practice neither was needed. Linux already had `build-linux.sh`, which is CMake-based and complete; the reason it had been left behind was an assumption that it needed Unity's PluginAPI headers, and it does not, because Linux resolves to the stub backend and the include is gated off. It builds in a plain container. Windows ARM64 needed the CMakeLists repair described above rather than a script of its own.


towneh:
PR remains in draft status until I can complete a new round of Windows stand-alone and Quest Pro regression tests tomorrow.

towneh:
This was a journey. Simplified the Android path to use the same portable demuxer as Windows.

PR #1032, opened 2026-09-01
AI disclosure edits

yewnyx:
## Summary
Some AI disclosure tweaks for voice and consistency, some for clarification/suggestion

## Required checks
All N/A, prose edit only

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -6,17 +6,17 @@ Basis is developed with help from AI tools, and we'd rather say so plainly than 
 
 AI coding assistants (LLM-based tools such as Claude Code) are used in Basis development for code only: implementation, refactoring, tests, debugging, and developer docs. AI is not used for creative content: the art, imagery, audio, and 3D assets in Basis are made by humans.
 
-As a concrete example, AI wrote many of the automated tests in Basis. That work would have taken a human months, and it means far more of the codebase is covered by tests than would otherwise be possible.
+As a concrete example, AI was used to wrote many of the automated tests in Basis. Without AI assistance, a human alone would have taken months; with AI assistance, far more of the codebase is covered by tests than would have otherwise been practical.
 
-Direction, review, and responsibility stay human. Changes land because a maintainer decided they should, reviewed them, and stands behind them, and the quality bar is the same for every change regardless of how it was written. We don't tag individual commits or files as AI-assisted; this document is the project-level disclosure instead.
+Unmoved by that assistance is the human element of direction, review, and responsibility. Changes land because contributors and maintainers decide they should, review them, and stand behind them, and the quality bar is the same for every change regardless of how it was written. We don't tag individual commits or files as AI-assisted; this document is the project-level disclosure instead.
 
 ## What ships in Basis
 
-**Basis does not embed generative AI.** The client and server do not call LLM or other generative AI services, and Basis does not send user data (voice, motion, text, or anything else) to AI services or collect it to train AI models.
+**Basis does not embed generative AI.** The client and server do not invoke LLMs or other generative AI services, and Basis does not send user data (voice, motion, text, or anything else) to AI services or collect it to train AI models.
 
 That's a commitment about the future, not just a description of today: **we will never include generative image or video AI in Basis.**
 
-Basis does ship small machine-learning models for classic signal-processing jobs, running entirely on your device, although these are very different from LLMs or generative ai.
+Basis does ship small machine-learning models for classic signal-processing jobs, running entirely on your device, although these are very different from LLMs or generative AI.
 
 - **RNNoise**: neural noise suppression on your own microphone input.
 - **OpenLipSync**: lip-sync viseme inference from your own voice, via ONNX Runtime.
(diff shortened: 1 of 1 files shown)


AI.md
```diff
@@ -6,33 +6,35 @@ Basis is developed with help from AI tools, and we'd rather say so plainly than
 
 AI coding assistants (LLM-based tools such as Claude Code) are used in Basis development for code only: implementation, refactoring, tests, debugging, and developer docs. AI is not used for creative content: the art, imagery, audio, and 3D assets in Basis are made by humans.
 
-As a concrete example, AI wrote many of the automated tests in Basis. That work would have taken a human months, and it means far more of the codebase is covered by tests than would otherwise be possible.
+As a concrete example, AI was used to wrote many of the automated tests in Basis. Without AI assistance, a human alone would have taken months; with AI assistance, far more of the codebase is covered by tests than would have otherwise been practical.
```
Toys0125: Ai was used to write not wrote.
IMHO I would remove the ; and make it a period.

AI.md
```diff
@@ -6,33 +6,35 @@ Basis is developed with help from AI tools, and we'd rather say so plainly than
...
 
-Direction, review, and responsibility stay human. Changes land because a maintainer decided they should, reviewed them, and stands behind them, and the quality bar is the same for every change regardless of how it was written. We don't tag individual commits or files as AI-assisted; this document is the project-level disclosure instead.
+Unmoved by that assistance is the human element of direction, review, and responsibility. Changes land because contributors and maintainers decide they should, review them, and stand behind them, and the quality bar is the same for every change regardless of how it was written. We don't tag individual commits or files as AI-assisted; this document is the project-level disclosure instead.
```
Toys0125, line 11: The first sentence sounds like it came from an AI. Maybe "People must control the direction....

yewnyx: Nope, human. Though people have told me I write like AI, all my modifications to the doc are purely human

AI.md
yewnyx: whoops. will fix in next pass

AI.md
```diff
@@ -6,33 +6,35 @@ Basis is developed with help from AI tools, and we'd rather say so plainly than
...
+
 ## AI-assisted contributions
 
 AI-assisted contributions are permitted, and the same requirements as [CONTRIBUTING.md](./CONTRIBUTING.md) apply. There are, however, some additional considerations to keep in mind for AI use.
 
 - **You are the author.** Understand what you submit well enough to explain and defend it in review. "The AI wrote it" is not an answer to a review question.
-- **Code within your limits.** Don't submit code that you couldn't have written by hand given enough time. AI should automate the time consuming tasks but please don't use it as a replacement for understanding what's going on in your code. You need to be firmly at the wheel.
+- **Code within your limits.** You may be tempted to submit code that you couldn't have written by hand given enough time. Do not use AI as a crutch to replace your understanding of what your code is doing. You need to be firmly at the wheel.
```
Toys0125: In a contribution guide using the word tempted is frowned upon in most communities i talked to. Using don't was more concrete.

AI.md
yewnyx: The semicolon is warranted because of the parallel structure in the first half and the second half - the semicolon binds it together more closely than separate sentences would

AI.md
yewnyx: I'll have to think about an alternative wording.

The problem I have with the prose pre-edit is that people will get Dunning-Kruger'ed into convincing themselves that the stuff they submitted is something they could have written, that just because they understand the output, they could have arrived there. So it becomes a line that applies to people who are conscientious and not to those who aren't, which is something of an insidious flaw.

AI.md
```diff
@@ -6,33 +6,35 @@ Basis is developed with help from AI tools, and we'd rather say so plainly than
...
-- **Don't submit AI generated documentation.** Many projects use agentic frameworks like "superpowers" and similar and produce large amounts of AI generated markdown files to help maintain history and context. However in practice this creates a large amount of context bloat for the agent, and also tends to get stale and out of sync with the code. If your PR includes a markdown file, be sure that you wrote it yourself and took the time to understand it. No one wants to read a wall of AI generated text.
+- **Documentation also requires maintenance.** Submitting an intense amount of prose subjects the project to an intense amount of pressure to review, correct, and maintain it. In practice, this requires a lot of human oversight to prevent it from becoming stale, in addition to any tokens spent to create it, and context bloat from ingesting it. Contributing documentation must be done in a way that is intelligible to humans, and maintainable on a reasonable scale. Endless walls of AI-generated text are not a net positive for the project.
```
Toys0125, line 36: I believe you lost the intent the original message was saying that AI assistance tools like to use Markdown files and tend to add context bloat and stale documenting as time goes on. Its always better to have them read the official Docs since we humans actually have to read them and update them.

Atm you wrote the exact saying of "Hey keep docs updated because they become stale." Its not this repo for docs to be handled.

AI.md
Toys0125: Okay I know you wrote it, but to the average person/ new contributor, (Who doesnt know you) do you want them to think all the files here are written by AI?

AI.md
yewnyx: I countermanded the intent of the original message because I don't agree with the thrust of it. On some private personal projects I spend a lot of tokens on extensive documentation even if I don't directly consume it, because my approach to it has significantly reduced context bloat, in exchange for being spendy on the tokens every so often. By paying upfront for it and cross-referencing relevant documents, the AI ends up getting stuff right on a first pass without re-ingesting the whole project or even the whole of the documentation. But it's way more than a human would want to read, and changes often.

So the message itself is, in my view, factually incorrect in my experience, but directionally correct in terms of "spamming docs sucks and hurts the project".

AI.md
yewnyx: I want them to think that a human will take responsibility for it either way. And if it's literally in my voice, typed by these hands, and they pick that out, it's a good moment to have a conversation about that.

AI.md
yewnyx: edited in latest

